PR #120, opened 2026-03-04
feat: compaction reset — trigger session reset after compaction with BOOT.md injection

MrGPUs:
## Problem

When OpenClaw compacts a conversation (trimming old messages to stay within context limits), the agent continues in a degraded state — it has the compacted context but no clean recovery path. The agent doesn't know a compaction happened, can't load its working state, and may lose track of what it was doing.

## Solution

When `compactionResetEnabled` is true (opt-in, default false), after compaction:

1. Call `api.resetSession()` to start a fresh session
2. Write a per-session signal file with compaction metadata (only after reset succeeds)
3. On next `recall()`, detect the signal file and inject BOOT.md content + compaction context
4. Consume the signal file (one-shot injection)

This runs alongside the existing checkpoint recovery. Checkpoint provides raw recent turns; BOOT.md provides high-level working state.

## Multi-agent safety

Every finding from the Codex automated review of PR #117 (now closed) has been addressed:

| Finding | Severity | Fix |
|---------|----------|-----|
| Shared mutable `_recallWorkspaceOverride` race condition | P1 | Per-session `Map<string, string>` keyed by sessionKey |
| No session validation on signal consumption | P1 | Signal `sessionKey` validated before injection |
| Single signal file per workspace (multi-session overwrite) | P1 | Per-session signal files: `.compaction-reset-signal-{sessionKey}` |
| Signal written before reset succeeds | P2 | Signal written only after `api.resetSession()` returns `ok: true` |
| Workspace overrides leak memory when feature is off | P2 | Map entries cleaned up unconditionally in every `recall()` call |

## Stale signal cleanup

On `initialize()`, the orchestrator sweeps the workspace for signal files older than 1 hour and removes them. Prevents orphaned signals when agents are removed from config.

## Registration model change

The gateway calls `register()` once per agent. Previously, a globalThis guard skipped all subsequent registrations, leaving later agents with zero hooks. Now:
- **Hooks**: re-register for each registry
- **Tools, CLI, service**: register only once (global registries)
- **Orchestrator**: singleton, shared across all registrations

## Config

```json
{ "compactionResetEnabled": true }
```

Default: `false` (opt-in).

## Dependencies

Requires OC fork with `api.resetSession()` (PR #29985). Degrades gracefully with an error log if unavailable. Note: `resetSession` is accessed via `api as any` cast since `OpenClawPluginApi` type doesn't include it yet.

## Testing

9 new tests (`tests/compaction-reset.test.ts`):
- Signal file per-session naming and round-trip
- Per-session signal isolation (multi-agent)
- Stale signal detection by file age
- Cross-session sessionKey validation
- BOOT.md injection when present
- BOOT.md missing produces warning
- Workspace override Map per-session cleanup
- Map cleanup runs even when feature disabled
- Startup sweep removes stale signals, preserves fresh ones

All 680 tests pass (671 existing + 9 new), 0 failures. Typecheck clean. Build clean.

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -719,6 +719,11 @@
         "default": 15,
         "description": "Number of turns per checkpoint"
       },
+      "compactionResetEnabled": {
+        "type": "boolean",
+        "default": false,
+        "description": "Trigger session reset after compaction with BOOT.md injection (requires OC fork with api.resetSession)"
+      },
       "hourlySummariesEnabled": {
         "type": "boolean",
         "default": true,
@@ -1818,6 +1823,11 @@
       "label": "Checkpoint Turn Interval",
       "advanced": true,
       "placeholder": "15"
+    },
+    "compactionResetEnabled": {
+      "label": "Compaction Reset",
+      "help": "Reset session after compaction and inject BOOT.md for recovery",
+      "advanced": true
     },
     "hourlySummariesEnabled": {
       "label": "Enable Hourly Summaries",
--- a/src/orchestrator.ts
+++ b/src/orchestrator.ts
@@ -2814,9 +2814,9 @@ export class Orchestrator {
     if (transcriptSection) {
       this.appendRecallSection(sectionBuckets, "transcript", transcriptSection);
     }
-    // Compaction reset context — injected after transcript (additive to checkpoint/transcript).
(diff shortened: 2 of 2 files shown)


MrGPUs:
@codex review

Issue #403, opened 2026-04-11
OpenClaw loads plugin, but before_prompt_build/agent_end never fire on 2026.4.11


tobalsan:
Hey, facing a weird one. I spent 8 hours trying to figure this out but... to no avail.

I got Remnic installed on OpenClaw, but memory extraction never kicks in.

At first I thought this was just a Remnic config / packaging issue.
But after digging, it looks more like an OpenClaw hook problem.

What I verified:

- OpenClaw: `2026.4.11`
- Remnic plugin: `@remnic/plugin-openclaw@1.0.3`
- yes, I had to **manually patch** the package to install because the published package depends on `"@remnic/core": "workspace:^"`
- plugin loads fine after patching
- `openclaw plugins info openclaw-engram` shows typed hooks including `before_prompt_build` and `agent_end`
- Remnic init logs appear normally

Then I made a tiny separate plugin that does nothing except:

- `api.on("before_prompt_build", ...)`
- `api.on("agent_end", ...)`
- append to a local log file

That plugin also loads fine.
But on real agent runs, **neither hook fires at all**.
Only the plugin registration logs appear.

So at least on my setup, this seems bigger than Remnic itself.
It looks like OpenClaw is loading typed hooks but not actually invoking them.

A few OpenClaw issues seem related:

- `openclaw/openclaw#5513`
- `openclaw/openclaw#52810`
- `openclaw/openclaw#21862`

What’s confusing me is: you don’t seem to be hitting this locally, otherwise the OpenClaw plugin would be dead on arrival.

So I’m wondering:

- have you seen this on `2026.4.11`?
- is there some known OpenClaw build / runtime path where these hooks work and others where they silently don’t?
- are you relying on a different path now for capture, or still `agent_end`?
- anything obvious I should compare in my setup?

Happy to paste the minimal repro plugin if useful. It's tiny.

joshuaswarren:
Going to take 2 PR's to fix this, two separate bugs. PR #404 resolved the first half. Working on the 2nd half now.

joshuaswarren:
@tobalsan - both issues you reported here should be resolved now. 

I _think_ the root cause was the fact that when I renamed the memory plugin from Engram to Remnic, it was no longer being registered in OpenClaw under the right name as a memory plugin. Memory plugins are automatically inserted into the turns in the right places. 

What to do:
[...]

PR #405, opened 2026-04-11
feat: rename plugin id to openclaw-remnic (#403, 1/2)

joshuaswarren:
## Summary

Implements option C from the #403 planning discussion: split the plugin id between the two published packages.

- `@remnic/plugin-openclaw` (and the root dev manifest): id **`openclaw-remnic`** — forward-looking, matches the package name
- `@joshuaswarren/openclaw-engram` shim: id stays **`openclaw-engram`** — this package exists as a legacy compat alias, keeping the old id is the whole point

### What changed

- **Manifest id flips** — root `openclaw.plugin.json` and `packages/plugin-openclaw/openclaw.plugin.json` now declare `"id": "openclaw-remnic"`. The shim manifest is untouched.
- **`PLUGIN_ID` / `LEGACY_PLUGIN_ID` constants** — new `packages/remnic-core/src/plugin-id.ts` exports both constants; re-exported from `@remnic/core`. Internal code now references constants instead of hardcoded strings.
- **Config lookup backwards compat** — `access-cli.ts`, `operator-toolkit.ts`, and `src/index.ts` now try `PLUGIN_ID` first, fall back to `LEGACY_PLUGIN_ID`, so users with existing `openclaw.json` using the old key keep working until they update.
- **QMD collection names preserved** — the default collection names (`openclaw-engram`, `openclaw-engram-cold`, `openclaw-engram-conversations`) are kept as-is with `TODO(#403)` comments. This means existing installs don't lose vector store data on upgrade. Users who set `qmdCollection` explicitly are unaffected.
- **`sample-openclaw-config.json`** updated to the canonical example with `plugins.slots.memory: "openclaw-remnic"` and `plugins.entries.openclaw-remnic`.
- **New regression test** `tests/plugin-id-split.test.ts` locks the id split in place — asserts root+plugin-openclaw use `PLUGIN_ID`, shim uses `LEGACY_PLUGIN_ID`, and they're distinct.
- **Updated tests** in `register-multi-registry.test.ts` and `operator-toolkit.test.ts` updated to assert `openclaw-remnic`.
- **compat/checks.ts** updated to accept both ids as valid (so the shim and the new package both pass the compat check).

Closes part of #403. A follow-up PR (#403 part 2) will add `remnic openclaw install` CLI, README/install docs updates, a startup log line, and the `remnic doctor` extensions.

## Test plan

- [x] `pnpm run check-types` — passes
- [x] `pnpm run build` — passes
- [x] `tests/plugin-id-split.test.ts` — 6 new tests pass (locked regression)
- [x] `tests/compat-checks.test.ts` — 8 tests pass (both ids accepted)
- [x] `tests/operator-toolkit.test.ts` — 20 tests pass
- [x] `tests/cli-compat.test.ts` — 3 tests pass
- [x] `tests/monorepo-structure.test.ts` — passes
- [ ] Full `pnpm test` — pre-existing timeout in CI (unrelated to this PR; reproduces on clean `main`)

## PR checklist

- [x] All new logic is covered by tests (plugin-id-split regression test)
- [x] Pre-commit hooks pass locally
- [x] Docs updated (sample-openclaw-config.json)
- [x] No secrets or creds committed
- [x] PR diff is focused and atomic

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it changes plugin identity/config resolution and reworks `globalThis`-scoped runtime singletons; mis-keying or resolver bugs could route reads/writes to the wrong memoryDir or duplicate CLI/service registration during migration.
> 
> **Overview**
> Renames the canonical OpenClaw plugin id to `openclaw-remnic` (root + `@remnic/plugin-openclaw` manifests, sample config, tests) while keeping the legacy shim package id as `openclaw-engram` for backward compatibility.
> 
> Centralizes config entry resolution into new `@remnic/core` exports `PLUGIN_ID`, `LEGACY_PLUGIN_ID`, and `resolveRemnicPluginEntry`, then updates all config-loader call sites (core access CLI, operator toolkit, codex materialize, plugin register/hook policy reads) to use consistent slot/preferred-id fallbacks; the shim `engram-access` now passes `preferredId: "openclaw-engram"` to avoid targeting the canonical block.
> 
> Scopes `@remnic/plugin-openclaw` runtime singletons/guards on `globalThis` by `serviceId` so canonical + shim can coexist in one process without sharing orchestrators/state, while keeping an unkeyed `__openclawEngramOrchestrator` mirror for observers. Adds regression tests for id split, entry resolution, and multi-id same-process behavior, and updates compat checks to accept both manifest ids.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1ec7297dda7eed3346f7f6d8b34c71e96b442924. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joshuaswarren (commented):
Fixed in commit 33355ca:  now reads the source manifest, patches the id back to , and writes it — instead of a raw . The shim manifest will always retain the legacy id regardless of what plugin-openclaw declares.

joshuaswarren:
@cursor review

joshuaswarren:
Addressed both unresolved review threads in fe11b8c:

**Cursor Medium — `operator-toolkit.ts:379` — Missing slot-based config lookup**
`loadCliPluginConfig` now resolves the active config entry via `plugins.slots.memory` first, mirroring the pattern in `access-cli.ts`, `loadPluginEntryFromFile`, and `readPluginHooksPolicy`. Migration configs with both `openclaw-remnic` and `openclaw-engram` entries now honour whichever id the operator configured in the memory slot, instead of always preferring the new id.

**Codex P1 — `src/index.ts:179` — Keep shim-exported plugin entry on legacy id**
Fixed at the shim layer (`packages/shim-openclaw-engram/src/index.ts`) rather than in `src/index.ts`, since the shim *is* the legacy alias: the shim now wraps the re-exported default plugin definition with `id: "openclaw-engram"`, so runtime readers of `pluginDefinition.id` agree with the shim's manifest declaration. The id is hardcoded (not imported from `@remnic/core`) because the shim's identity *is* the legacy string.

Verified locally:
- `pnpm run check-types` (root): clean
- `pnpm --filter @remnic/core run check-types`: clean
- `pnpm --filter @joshuaswarren/openclaw-engram run check-types`: clean
- Targeted tests: `plugin-id-split.test.ts`, `shim-openclaw-engram-package.test.ts`, `monorepo-structure.test.ts` (16/16 pass), `operator-toolkit.test.ts`, `operator-toolkit-cli.test.ts` (27/27 pass).

joshuaswarren:
Addressed three Cursor Bugbot findings from the latest push:

**Codex P1 — `packages/plugin-openclaw/openclaw.plugin.json:2` — Update downstream config loaders for renamed plugin id (commit d0bfdf0)**
`unwrapOpenClawEntry` in `packages/plugin-codex/bin/materialize.cjs` now tries `plugins.slots.memory` (the operator-configured active entry) first, then `openclaw-remnic`, then falls back to `openclaw-engram`. Same slot-based fallback pattern used in `access-cli.ts`, `loadPluginEntryFromFile`, and `loadCliPluginConfig`.

**Cursor Medium — `packages/shim-openclaw-engram/src/index.ts:17` — Shim registers service with new id, not legacy id (commit 8ffb9ca)**
`register()` in `src/index.ts` now captures `this.id` at the start of the method and uses that as the service id passed to `api.registerService()`. When the backward-compat shim re-exports with `id: "openclaw-engram"` overriding the spread, the service is now registered under the correct legacy id, preserving backwards-compat.

**Cursor Low — `src/index.ts:83` — Slot lookup uses inconsistent early-return vs nullish-coalescing (commit 8ffb9ca)**
`loadPluginEntryFromFile` in `src/index.ts` now uses the ternary-plus-`??` pattern, matching `access-cli.ts` and `operator-toolkit.ts`. The behaviour is semantically identical but now consistent across all three entry-point loaders.

joshuaswarren:
**Codex P2 — `src/index.ts:83` — Restrict slot-based config selection to Remnic plugin IDs (commit 51157f0)**

`loadPluginEntryFromFile()`, `loadPluginConfig()` (access-cli.ts), and `loadCliPluginConfig()` (operator-toolkit.ts) now guard the `plugins.slots.memory` slot lookup to only honour slots that point to a known Remnic plugin id (`openclaw-remnic` or `openclaw-engram`). Previously, if `plugins.slots.memory` pointed to a different memory plugin (e.g. a third-party plugin in a mixed-plugin install), its config block could be silently loaded into Remnic. All three entry-point loaders are updated together so they continue to agree on the selection algorithm.

joshuaswarren:
Two more Codex P2 findings addressed in commit 64cffca:

**PRRT_kwDORJXyws56U-f0 — `readPluginHooksPolicy()` in `src/index.ts:118`**
Added the `isRemnicSlot` guard (matching the pattern from `loadPluginEntryFromFile`) so the `api.config.plugins.slots.memory` slot is only trusted when it equals `openclaw-remnic` or `openclaw-engram`. Foreign memory plugin slots no longer cause Remnic to read hook policies from an unrelated plugin entry.

**PRRT_kwDORJXyws56U-f2 — `unwrapOpenClawEntry()` in `materialize.cjs:99`**
Validated `plugins.slots.memory` before including it in `candidateIds` — only includes it when it is one of the known Remnic plugin IDs. Foreign slot IDs are now silently skipped and the function falls through to the hardcoded `openclaw-remnic`/`openclaw-engram` candidates.

joshuaswarren:
Cursor Bugbot medium finding addressed in commit c404a2f:

**Slot-based config resolution duplicated across five locations**

Extracted the slot → `PLUGIN_ID` → `LEGACY_PLUGIN_ID` lookup logic into a single `resolveRemnicPluginEntry` function exported from `@remnic/core` (in `packages/remnic-core/src/plugin-id.ts`). All five loader sites now delegate to this shared helper:
- `loadPluginEntryFromFile` in `src/index.ts`
- `readPluginHooksPolicy` in `src/index.ts`
- `loadPluginConfig` in `access-cli.ts`
- `loadCliPluginConfig` in `operator-toolkit.ts`
- `loadRawConfig` in `materialize.cjs` (resolves via dynamic import of `@remnic/core`)

The `materialize.cjs` approach passes the resolver function as a parameter so the CJS file avoids replicating ESM-only module logic while still staying synchronous for file I/O.

joshuaswarren:
Two Cursor Bugbot findings addressed in commit 95f5af9:

**High — Legacy flat-config fallback lost in materialize.cjs (PRRT_kwDORJXyws56VBH3)**
`loadRawConfig` in `materialize.cjs` was skipping flat legacy config files (e.g. `~/.config/remnic/config.json`) because `resolveRemnicPluginEntry` returns `undefined` for configs without a `plugins.entries` subtree. Restored: when `resolveEntry` returns `undefined` and the file has no `plugins` key, treat the raw object as the config (matching the previous `unwrapOpenClawEntry` fallback behavior).

**P2 — loadPluginEntryFromFile fallback order ignores active plugin id (PRRT_kwDORJXyws56U_5J)**
When `plugins.slots.memory` is absent and a migration config has both `openclaw-remnic` and `openclaw-engram` entries, shim installs (`id="openclaw-engram"`) would pick the canonical entry. Fix: `resolveRemnicPluginEntry` now accepts an optional `preferredId` parameter that is tried before the hardcoded fallbacks when no slot override is set. `serviceId` is threaded through `loadPluginConfigFromFile` and `readPluginHooksPolicy` so shim installs consistently prefer their own entry.

joshuaswarren:
## Codex P1 addressed in 29ed1b7

Thanks @chatgpt-codex-connector for catching this — fresh evidence was accurate:
`loadPluginConfig()` in `access-cli.ts` was calling `resolveRemnicPluginEntry(raw)`
without a `preferredId`, so shim-CLI (`@joshuaswarren/openclaw-engram`) users could
silently target the canonical `"openclaw-remnic"` memory store instead of their own
legacy block.

### Fix

Thread `preferredId` through the full CLI chain:

```
runCli({ preferredId }) → main(argv, options) → buildRuntime(preferredId)
  → loadPluginConfig(preferredId) → resolveRemnicPluginEntry(raw, preferredId)
```

The shim bin wrapper now hardcodes `preferredId: "openclaw-engram"`
(`packages/shim-openclaw-engram/bin/engram-access.js`) — the id is literal
(not imported from `@remnic/core`) because the shim *is* the legacy alias.

### Regression coverage

New `tests/resolve-remnic-plugin-entry.test.ts` locks the full resolution matrix
in place — **10 tests pass**:

| Scenario                                                              | Expected    |
| --------------------------------------------------------------------- | ----------- |
| null/undefined/non-object input                                       | `undefined` |
| `plugins.entries` missing                                             | `undefined` |
| Both entries, no `preferredId`, no slot                               | canonical   |
| Both entries, `preferredId="openclaw-engram"`, no slot **(shim CLI)** | **legacy**  |
| Both entries, `preferredId="openclaw-remnic"`, no slot                | canonical   |
| Both entries, unknown `preferredId` (safety guard)                    | canonical   |
| Both entries, active slot on canonical, shim `preferredId`            | canonical   |
| Foreign slot (non-Remnic), shim `preferredId`                         | legacy      |
| Only legacy entry, no `preferredId`                                   | legacy      |
| Shim bin wrapper passes `preferredId: "openclaw-engram"` (static)     | ✓           |

### Verification

- `pnpm run check-types` (root) — ✅ pass
- `npx tsx --test tests/resolve-remnic-plugin-entry.test.ts` — ✅ 10/10 pass
- `npx tsx --test tests/plugin-id-split.test.ts tests/shim-openclaw-engram-package.test.ts tests/cli-access.test.ts` — ✅ 10/10 pass
- Changelog updated under `[Unreleased]`

joshuaswarren:
### P2 fix pushed — runtime singletons now scoped per `serviceId` (commit 09b84eb)

Addresses Codex P2 ([thread PRRT_kwDORJXyws56VHv0](https://github.com/joshuaswarren/remnic/pull/405#discussion_r-P2)) on `src/index.ts`.

**What changed:**
- All per-plugin runtime state on `globalThis` is now keyed by `::${serviceId}`:
  - `__openclawEngramRegistered::<id>`
  - `__openclawEngramHookApis::<id>`
  - `__openclawEngramOrchestrator::<id>`
  - `__openclawEngramAccessService::<id>`
  - `__openclawEngramAccessHttpServer::<id>`
  - `__openclawEngramServiceStarted::<id>`
  - `__openclawEngramInitPromise::<id>`
- `serviceId` is derived from `this.id` inside `register()`, so the canonical (`openclaw-remnic`) and legacy shim (`openclaw-engram`) plugin ids get independent runtime state when both load in one process (migration install).
- The unkeyed `__openclawEngramOrchestrator` slot is retained as a best-effort mirror for cross-plugin observers that do not know the `serviceId`. It points at whichever plugin registered most recently and is never read during per-service startup — all service code reads from the keyed slot.
- The migration promise (`__openclawEngramMigrationPromise`) stays unkeyed because the legacy-dir migration is a one-time process-wide operation.

**Regression test:**
A new test in `tests/register-multi-registry.test.ts` — `register() scopes runtime singletons per serviceId when two plugin ids share a process (regression: #403 P2)` — registers both plugin ids against separate api objects in the same process and asserts:
- both `__openclawEngramRegistered::<id>` guard slots are set independently
- canonical and legacy plugins get distinct orchestrator instances
- canonical and legacy plugins get distinct access service instances
- each plugin registers its own service id
- the unkeyed mirror points at the most-recently-registered plugin

The three sibling test files (`sdk-compat-hook-handlers.test.ts`, `sdk-compat-integration.test.ts`, `remnic-plugin-register-migration.test.ts`) were updated to clean/read the keyed slot names.

All four test files pass individually (typecheck clean, all individual test assertions pass).

@cursor please re-review.

joshuaswarren:
### Cursor Bugbot Low finding addressed in 4c8c5f1

**Finding**: `saveAndResetGlobals` and `restoreGlobals` didn't save/restore the unkeyed `__openclawEngramOrchestrator` mirror that `register()` now maintains — state could leak across test cases.

**Fix**: `saveAndResetGlobals`/`restoreGlobals` now cover both the keyed slot and the unkeyed mirror. Scenario C's ad-hoc save/restore of the unkeyed mirror has been removed since it's now redundant with the helper coverage. All 11 tests in `register-multi-registry.test.ts` still pass, including the new `#403 P2` regression test.

@cursor please re-review.

joshuaswarren:
Fixed in baff4cb — added `UNKEYED_ORCH_MIRROR_KEY` (`__openclawEngramOrchestrator`) to `resetGlobals()` in `tests/sdk-compat-integration.test.ts`, matching the cleanup already present in the other three test files. All 7 sdk-compat-integration tests pass; typecheck clean.

joshuaswarren:
### Codex P2 CLI dedupe guard fixed in bd4eb09

**Finding**: `REGISTERED_GUARD` keyed per-serviceId caused both `openclaw-remnic` and `openclaw-engram` to independently call `registerCli()` in migration installs, creating duplicate command trees.

**Fix**: Introduced a separate process-global `CLI_REGISTERED_GUARD` (`__openclawEngramCliRegistered`) that ensures CLI registration happens exactly once regardless of plugin id count. `stop()` clears it on full stop (same policy as `REGISTERED_GUARD`).

**Test**: Scenario C now asserts `canonical.getCliCount() > 0` and `legacy.getCliCount() === 0`. All 11 multi-registry tests pass, all 7 sdk-compat-integration tests pass, typecheck clean.

joshuaswarren:
### CLI guard stop() fix in 682eb26

**Codex P2** (PRRT_kwDORJXyws56WHTe) + **Cursor Medium** (PRRT_kwDORJXyws56WHwy): Both flagged the same issue — `stop()` clearing `CLI_REGISTERED_GUARD` allowed the still-running plugin in a migration install to re-register CLI and create duplicate command trees.

**Fix**: Removed the `CLI_REGISTERED_GUARD = false` from stop(). CLI commands are never unregistered from the gateway's central registry on stop(), so the guard must persist. It naturally resets on full process restart when globalThis is cleared.

All 11 register-multi-registry tests pass, typecheck clean.

joshuaswarren:
### CLI guard refcount fix in 9d34209

**Cursor Medium** (PRRT_kwDORJXyws56WJM2): Previous commit never cleared CLI_REGISTERED_GUARD in stop(), risking silent CLI loss after gateway reload.

**Fix**: Introduced `CLI_ACTIVE_SERVICE_COUNT` — a process-global refcount incremented in start(), decremented in stop(). CLI_REGISTERED_GUARD is only cleared when the count drops to zero (all Remnic services stopped), so:
- One plugin stopping in a multi-plugin install does NOT clear the guard
- A true full stop (all services down) DOES clear it for reload re-registration

All 11 register-multi-registry + 7 sdk-compat-integration tests pass; typecheck clean.

joshuaswarren:
Fixed in 1ec7297 — guarded CLI refcount decrement in start() catch block with `if (didCountStart)` check. Prevents double-decrement when stop() runs during in-flight init.

PR #503, opened 2026-04-19
Add OpenClaw memory runtime capability shim

earlvanze:
## Summary
- register a `MemoryPluginRuntime` alongside Remnic's unified memory capability
- expose a lightweight search manager shim so OpenClaw doctor/status can see the active backend instead of treating Remnic as unregistered
- report the active Remnic QMD command through `resolveMemoryBackendConfig()` and keep file reads constrained to workspace/memory roots

## Why
Recent OpenClaw builds probe the active memory plugin through the unified capability runtime hook. Remnic already registers `promptBuilder` and `publicArtifacts`, but it does not currently provide `capability.runtime`, so `openclaw doctor` reports:

- `No active memory plugin is registered for the current config.`

...even while Remnic is the selected memory-slot plugin and memory status/search are otherwise working.

## Notes
- This is a compatibility shim, not a new retrieval path.
- The runtime manager delegates to existing Remnic orchestrator/search state.
- `readFile()` is restricted to the workspace/memory roots to avoid widening file access.

## Validation
- patched the installed OpenClaw Remnic extension locally with the same runtime shim
- `node --check /home/umbrel/.openclaw/extensions/openclaw-remnic/dist/index.js`


joshuaswarren:
Reviewed and confirmed this will work for both legacy users running Remnic directly in OpenClaw and those using the new plugin approach. Once the Codex feedback is addressed this PR is ready to merge. Thanks for contributing it!

Issue #459, opened 2026-04-18
feat: export Remnic memories as WeClone fine-tuning datasets


joshuaswarren:
## Summary

Add the ability to export Remnic's structured memory extractions as high-quality fine-tuning datasets in [WeClone](https://github.com/xming521/weclone)-compatible format, producing cleaner training data than raw chat logs.

## Motivation

WeClone trains on raw chat exports (Telegram JSON, etc.), which are noisy:
- Off-topic messages, spam, reactions, one-word replies
- PII that must be scrubbed (WeClone uses Presidio, but noise remains)
- No semantic structure — just chronological message dumps

Remnic already distills conversations into structured artifacts:
- **Facts**: durable knowledge with confidence scores and categories
- **Entities**: people, projects, tools with relationship graphs
- **Topics**: extracted themes with temporal context
- **Preferences**: user communication patterns and style signals

These are a much higher signal-to-noise source for fine-tuning a digital avatar. A user running Remnic alongside their daily AI tools accumulates exactly the kind of data WeClone needs — already categorized, deduplicated, and scored for importance.

### Package boundary

| Layer | Package | Responsibility |
|-------|---------|---------------|
| Core | `@remnic/core` | Generic training-data export pipeline (format-agnostic) |
| Adapter | `@remnic/export-weclone` (new) | WeClone-specific format adapter, dataset builder |

### Core additions (minimal)

Add a generic export interface to `@remnic/core` that any format adapter can implement:

```typescript
// In @remnic/core — generic export contract
interface TrainingExportOptions {
  memoryDir: string;
  since?: Date;
  until?: Date;
  minConfidence?: number;
  categories?: string[];
  includeEntities?: boolean;
  includeTopics?: boolean;
}

interface TrainingExportRecord {
  instruction: string;     // The "question" or prompt
  input: string;           // Additional context (optional)
  output: string;          // The "answer" — synthesized from memories
  category?: string;       // Memory category for filtering
  confidence?: number;     // Source memory confidence score
  sourceIds?: string[];    // Provenance back to original memories
}

interface TrainingExportAdapter {
  name: string;
  formatRecords(records: TrainingExportRecord[]): string;
  fileExtension: string;
}
```

This goes in core because other export targets (Axolotl, MLX, etc.) would use the same interface.

### WeClone adapter package

```
packages/export-weclone/
├── package.json
├── src/
│   ├── index.ts              # Package entry
[...]

PR #514, opened 2026-04-19
fix(export-weclone): harden training:export flag parsing (follow-up to #509)

joshuaswarren:
## Summary

Follow-up to #509 addressing two AI review comments that were posted after the fix-forward push but before the squash-merge landed.

- **Codex P2**: `--memory-dir --since 2026-01-01` silently dropped the memoryDir flag (the next token looked like a flag, so `resolveFlagStrict` returned undefined) and fell back to the resolver default. That is risky for a command that emits shareable datasets where a broadened filter can leak memories the user intended to exclude. Added `resolveRequiredValueFlag` that throws when a value-taking flag is present without a value, and routed every value-taking option through it: `--format`, `--output`, `--out`, `--memory-dir`, `--since`, `--until`, `--min-confidence`, `--categories`, `--max-pairs-per-record`.
- **Cursor (low severity)**: `isCalendarDateValid` was accidentally re-exported from `@remnic/core`'s training-export barrel. It was never a public API and has no external callers — un-exported to avoid committing to long-term support for an internal helper.

No behaviour change for well-formed invocations — only malformed flag sequences now fail fast instead of silently broadening the export.

## Test plan
- [x] `pnpm run check-types` clean
- [x] `tsx --test packages/export-weclone/src/*.test.ts` — 61/61 passing (adapter, privacy, style-extractor, synthesizer, integration).

Refs: #459, #509.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

---

> [!NOTE]
> **Medium Risk**
> Tightens CLI parsing for `training:export` to throw on previously-accepted malformed flag sequences, which could break automation but reduces risk of unintentionally exporting broader datasets. Also changes the public `@remnic/core` barrel exports, which may affect any downstream code importing the removed symbol.
> 
> **Overview**
> **Hardens `remnic training:export` argument parsing** by introducing `resolveRequiredValueFlag` and routing all value-taking flags (`--format`, `--output/--out`, `--memory-dir`, date filters, and numeric/category options) through it so a present-but-missing value now throws instead of silently defaulting.
> 
> **Narrows the `@remnic/core` training-export public API** by no longer re-exporting the internal `isCalendarDateValid` helper; only `parseStrictCliDate` remains exported from the barrel.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 423182e643f877540623a6d2beec7fc3703bfe62. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/remnic-cli/src/index.ts
+++ b/packages/remnic-cli/src/index.ts
@@ -3991,8 +3991,33 @@ interface ParsedTrainingExportArgs {
 }
 
 /**
- * Parse training:export CLI flags. Rejects unknown values instead of
- * silently defaulting, per CLAUDE.md #14/#51.
+ * Resolve a value-taking flag, rejecting the "flag present but missing
+ * value" case (e.g. `--memory-dir --since 2026-01-01`). CLAUDE.md #14
+ * requires that `--foo` without an argument throws rather than silently
+ * defaulting — critical here because `training:export` emits shareable
+ * data and a wrongly-broadened filter would leak memories the user
+ * intended to exclude (Codex review follow-up to PR #509).
+ *
+ * Returns `undefined` only when the flag is absent; throws when the flag
+ * is present but its value is missing or shaped like another flag.
+ */
+function resolveRequiredValueFlag(
+  args: string[],
+  flag: string,
+): string | undefined {
+  if (!hasFlag(args, flag)) return undefined;
+  const value = resolveFlagStrict(args, flag);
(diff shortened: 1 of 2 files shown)


joshuaswarren:
Bumping to wake AI reviewers; PR is small (2 files, ~40 lines) follow-up hardening for #509.

PR #541, opened 2026-04-19
refactor: extract shared abort-error helpers

joshuaswarren:
Resolves the Cursor low-severity finding from PR #533 that was deferred in favor of keeping slice 3's surface tight.

## Summary

The \`abortError\` / \`isAbortError\` / \`throwIfAborted\` pattern was independently implemented in three places:

- \`direct-answer-wiring.ts\` (simple throwIfAborted)
- \`harmonic-retrieval.ts\` (simple throwIfAborted)
- \`qmd.ts\` (full kit with message support)

Any future change to the abort-error convention had to hand-verify each call site. This PR introduces \`abort-error.ts\` consolidating all three helpers behind the qmd.ts-shaped contract.

- \`abortError(message): Error\` — Error with \`name === "AbortError"\`
- \`isAbortError(err): boolean\` — type-narrowing check
- \`throwIfAborted(signal, message?): void\` — default message \`"operation aborted"\`

## Changes

- New \`packages/remnic-core/src/abort-error.ts\` + 6 unit tests
- \`direct-answer-wiring.ts\`, \`harmonic-retrieval.ts\`, \`qmd.ts\` import from it and delete their local copies

## Compatibility

- Existing error shapes unchanged — still standard \`Error\` with \`name === "AbortError"\`.
- Local error messages differ slightly from before (e.g. \`"harmonic retrieval aborted"\` → \`"operation aborted"\`), but no caller parses the message; classification uses \`err.name\`.

## Test plan

- [x] \`abort-error.test.ts\` — 6/6 pass
- [x] \`direct-answer-wiring.test.ts\` — 15/15 pass (no regressions)
- [x] \`tsc --noEmit\` clean

Not a blocker for any #518 slice but makes future abort-related work easier.

---

> [!NOTE]
> **Low Risk**
> Low risk refactor that centralizes abort/cancellation error construction and checks; primary risk is subtle behavior/message differences affecting abort handling paths.
> 
> **Overview**
> Introduces a shared `abort-error.ts` module exporting `abortError`, `isAbortError`, and `throwIfAborted`, plus a new unit test suite covering their behavior.
> 
> Updates `direct-answer-wiring.ts`, `harmonic-retrieval.ts`, and `qmd.ts` to import these helpers and deletes their in-file implementations, standardizing abort error shape (`Error` with `name === "AbortError"`) and making abort checks use the shared `throwIfAborted` (with explicit per-module messages where needed).
> 
[...]

PR #552, opened 2026-04-19
fix(orchestrator): log extraction-queue aborts at debug, not error (#549)

### LINKED ISSUE #549: Extraction abort on session reset logged as error — should be debug/info level
## Description

When a session resets (e.g., user sends `/new`), Remnic sometimes logs this as an error:

```
[plugins] remnic: background extraction task failed — extraction aborted (before_extract)
```

This appears alongside a **successful** extraction that completes 4ms earlier:

```
17:20:55.502 remnic: persisted: 4 facts, 3 entities, 2 questions, 1 profile updates
17:20:55.506 remnic: background extraction task failed — extraction aborted (before_extract)
```

## Root Cause

When a session resets, two things happen concurrently:
1. The `before_reset` hook triggers a flush/extraction for the outgoing session
2. The new session starts and queues its own extraction

The second extraction finds the abort signal already set (from the session transition) and throws `extraction aborted (before_extract)` — which is caught at line ~32689 and logged as `log.error("background extraction task failed")`.

The abort is **working as designed** — it prevents duplicate extraction. But the log level and message make it look like a real failure rather than expected deduplication.

Additionally, the `before_reset flush timed out after 2000ms` warning often appears alongside this, because the flush races against the new session initialization.

## Code References

The error is logged here (`index.js`):
```js
// line ~32688
} catch (err) {
  log.error("background extraction task failed", err);
}
```

The abort is thrown here:
```js
// line ~32707
throwIfRecallAborted(options.abortSignal, "extraction aborted (before_extract)");
```

The before_reset flush timeout is here:
```js
// line ~58578
log.warn("before_reset flush timed out after ${cfg.beforeResetTimeoutMs}ms");
```

## Suggested Fix

1. **Differentiate expected aborts from real errors.** The `throwIfRecallAborted` path should not be logged at `error` level. An aborted extraction during session transition is normal, not a failure. Suggestion:
   ```js
   } catch (err) {
     if (err?.message?.includes("extraction aborted")) {
       log.debug("background extraction skipped — session transition abort", err);
     } else {
       log.error("background extraction task failed", err);
     }
   }
   ```

2. **Increase `beforeResetTimeoutMs`** (currently 2000ms) or make it configurable. On systems using a local LLM for extraction, the flush can take 2-5 seconds, making the 2s timeout too aggressive.

## Impact

- Spurious `error` log entries that confuse monitoring and suggest the extraction system is broken
- Operators waste time investigating non-issues
- Not a data loss problem — extraction works correctly, it just logs wrongfully

## Environment

- **OS:** Ubuntu 25.10 (Questing Quokka), kernel 6.17.0-22-generic
- **CPU:** 12th Gen Intel Core i7-12700H (14-core, 20-thread)
- **GPU:** NVIDIA GeForce RTX 3060 Laptop GPU, 6GB VRAM, driver 590.48.01, CUDA compute 8.6
- **RAM:** 30 GB DDR5
- **Node:** v24.14.0
- **Remnic:** v9.3.24 (openclaw-remnic plugin)
- **OpenClaw:** v2026.4.15
- **Local LLM:** llama.cpp v8642 (CUDA, RTX 3060 6GB)
- **Model:** gemma-4-E4B-it-Q3_K_S (Unsloth), served on port 8081 with -ngl 99 -c 32768
- **Ollama:** v0.6.14 on port 11434 (cloud model proxy)

joshuaswarren:
Closes #549.

## The bug

\`orchestrator.ts:8717\` (pre-fix line number) swallowed every queued-extraction error into \`log.error("background extraction task failed", err)\`. When a session reset fires the abort signal, any queued follow-up extraction throws \`extraction aborted (<stage>)\` via \`throwIfRecallAborted\` — intentional deduplication. The \`error\`-level log next to a successful \`persisted: N facts\` log misleads operators.

The reporter's example:

\`\`\`
17:20:55.502 remnic: persisted: 4 facts, 3 entities, 2 questions, 1 profile updates
17:20:55.506 remnic: background extraction task failed — extraction aborted (before_extract)
\`\`\`

## Fix

Gate the two queue catch sites on \`isAbortError(err)\` from the shared \`abort-error.ts\` module (added in PR #541).

- \`processQueue()\` per-task catch: AbortError → \`log.debug(...aborted (session transition))\`, real failures keep \`log.error\`.
- Outer \`processQueue().catch(...)\` follows the same pattern.

Discriminator is \`err.name === "AbortError"\` — the stable contract set by \`throwIfRecallAborted\` / \`abortError\`. The reporter's proposed \`err.message.includes("extraction aborted")\` string-match would be fragile: the message varies by stage (\`before_extract\`, \`before_clear_buffer\`, \`during_extract\`) and other abort sites use different messages entirely.

## Subsystem-hardening cleanup

The orchestrator still had its own private \`abortRecallError\` and \`throwIfRecallAborted\` helpers — noted but intentionally left by PR #541. Migrated them to thin wrappers over the shared \`abort-error.ts\`, preserving the \`"recall aborted"\` default message so the 9 call sites without an explicit message aren't disturbed. This ensures every abort in the orchestrator produces an error shape that \`isAbortError\` uniformly classifies. CLAUDE.md "Why Review Churn Happens" rule — wider the fix to the whole subsystem, not just the reported symptom.

## \`beforeResetTimeoutMs\` — already configurable

The reporter's second suggestion was to make \`beforeResetTimeoutMs\` configurable or raise the default. Verified it's already exposed in \`openclaw.plugin.json\` (default 2000ms, clamped \`[100, 30000]\`). Raising the default would slightly delay every session transition for every user, so this PR leaves the default alone and adds a doc note so local-LLM operators know to tune it.

## Tests

New: \`packages/remnic-core/src/orchestrator-extraction-queue.test.ts\` (4 cases)

- abort-error task → debug, zero error logs
- regular-error task → error log, unchanged
- mixed run (abort + failure) → one debug + one error
- \`queueProcessing\` flag clears on exit

Existing: \`orchestrator-flush.test.ts\` + \`abort-error.test.ts\` — 26/26 pass unchanged.

## Docs

\`docs/config-reference.md\` — new row for \`beforeResetTimeoutMs\` under Extraction Guardrails with tuning guidance for local-LLM operators and a cross-reference to #549.

Fixes #549.

---

> [!NOTE]
> **Low Risk**
> Low risk: changes are limited to log-level classification and shared abort helpers, with added tests to prevent regressions; functional extraction behavior should remain unchanged.
> 
> **Overview**
> Prevents spurious `error` logs during session resets by classifying extraction-queue `AbortError`s as intentional cancellations: both per-task failures in `processQueue()` and the outer `processQueue().catch(...)` path now log aborts at `debug` while keeping real failures at `error`.
> 
> Refactors orchestrator abort helpers to use the shared `abort-error.ts` utilities for consistent `AbortError` detection, adds a focused extraction-queue regression test suite for issue #549, and documents `beforeResetTimeoutMs` tuning guidance in `config-reference.md`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 9ab003cd0b4b63e3b990395be90b8681778ec269. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -83,6 +83,7 @@ Access-layer safety notes:
 | `extractionMaxEntitiesPerRun` | `6` | Cap on entities extracted per LLM call |
 | `extractionMaxQuestionsPerRun` | `3` | Cap on curiosity questions generated per LLM call |
 | `extractionMaxProfileUpdatesPerRun` | `4` | Cap on profile update statements per LLM call |
+| `beforeResetTimeoutMs` | `2000` | Max time (ms, clamped to `[100, 30000]`) to wait for a reset-triggered flush before returning control to the host. Operators running a local LLM for extraction often want this higher — a 7B model on CPU can take 2–5s per extraction, and the default can abort the queued follow-up flush before it completes. See issue #549 for the error-vs-debug log-level behavior around these aborts. |
 
 ## Search Backend (v9.0)
 
--- a/packages/remnic-core/src/orchestrator-extraction-queue.test.ts
+++ b/packages/remnic-core/src/orchestrator-extraction-queue.test.ts
@@ -128,3 +128,70 @@ test("processQueue clears queueProcessing on exit regardless of task outcomes", 
   await orch.processQueue();
   assert.equal(orch.queueProcessing, false);
 });
+
+// ── Outer processQueue().catch() branch (Codex follow-up on #549) ──────────
+
+type QueueProcessorOrchestrator = QueueOrchestrator & {
+  logExtractionQueueFailure: (err: unknown, source: "task" | "processor") => void;
+};
+
+test("logExtractionQueueFailure(processor) classifies AbortError as debug (#549)", async () => {
+  // Covers the outer `processQueue().catch(...)` path in
+  // `queueBufferedExtraction`.  A processor-level AbortError can
+  // bubble from e.g. a processQueue rewrite that awaits an external
+  // signal, so the handler must fail open to debug — otherwise a
+  // session-transition-triggered processor abort would get logged
+  // at error next to the successful extraction it just produced.
+  const { entries } = installCapturingLogger();
+  const orch = createQueueOrchestrator() as QueueProcessorOrchestrator;
+  orch.logExtractionQueueFailure(
+    abortError("queue processor aborted"),
+    "processor",
+  );
+  const errorEntries = entries.filter((e) => e.level === "error");
+  const debugEntries = entries.filter(
+    (e) =>
+      e.level === "debug" &&
+      e.message.includes("background extraction queue processor aborted"),
+  );
+  assert.equal(errorEntries.length, 0);
+  assert.equal(debugEntries.length, 1);
+});
+
+test("logExtractionQueueFailure(processor) preserves error-level for real failures", async () => {
+  const { entries } = installCapturingLogger();
+  const orch = createQueueOrchestrator() as QueueProcessorOrchestrator;
+  orch.logExtractionQueueFailure(
+    new Error("unexpected processor crash"),
+    "processor",
+  );
+  const errorEntries = entries.filter((e) => e.level === "error");
+  assert.equal(errorEntries.length, 1);
+  assert.ok(
+    errorEntries[0]?.message.includes(
+      "background extraction queue processor failed",
+    ),
+  );
+});
+
+test("logExtractionQueueFailure names the right layer (task vs processor)", async () => {
+  // Regression guard that keeps the two log messages distinct — a
+  // single shared message would lose the operator context that
+  // tells them whether a specific extraction aborted or the queue
+  // processor itself did.
+  const { entries } = installCapturingLogger();
+  const orch = createQueueOrchestrator() as QueueProcessorOrchestrator;
+  orch.logExtractionQueueFailure(abortError("a"), "task");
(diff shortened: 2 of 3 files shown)


joshuaswarren:
@chatgpt-codex-connector please review.

@cursor review

One-file fix for #549 — extraction-queue aborts during session transitions were logged at error; now they log at debug. Shared `isAbortError` is the discriminator so the AbortError contract stays unified across orchestrator / qmd / direct-answer-wiring / harmonic-retrieval. Migration of orchestrator's private abortRecallError/throwIfRecallAborted to thin wrappers over abort-error.ts closes the follow-up from PR #541. 4 new regression tests in orchestrator-extraction-queue.test.ts.

joshuaswarren:
@chatgpt-codex-connector thanks — good catch. Addressed in 9ab003cd:

- Extracted the classification logic into a single private `logExtractionQueueFailure(err, source: "task" | "processor")` helper. Both the per-task `processQueue()` catch and the outer `processQueue().catch(...)` now route through it.
- 3 new regression tests covering the processor path specifically: AbortError → debug, real error → error, and a layer-naming guard asserting task and processor aborts produce distinct log messages.
- 7/7 tests pass, tsc clean, no production behavior change.

@cursor review for the refactor.

PR #558, opened 2026-04-20
fix: harden local benchmark teardown and LM Studio diagnostics

joshuaswarren:
## Summary
- prevent local benchmark runs from hanging after successful completion by hardening bench adapter teardown and forcing the direct CLI entrypoint to exit cleanly
- preserve normal benchmark recall behavior for existing users instead of globally shrinking recall payloads
- add an explicit LM Studio context-window hint when a local OpenAI-compatible endpoint fails because the loaded model context is too small

## Why
The original local Gemma investigation surfaced two separate issues:
1. benchmark runs could finish but leave local resources open, which made the CLI appear hung
2. full `ama-bench` against LM Studio Gemma can fail with `n_keep >= n_ctx`, which is an LM Studio model context configuration issue rather than a Remnic benchmark logic bug

This PR fixes the first issue in code and makes the second issue diagnosable without changing benchmark semantics for everyone else.

## Verification
- `pnpm exec tsx --test packages/bench/src/adapters/remnic-adapter.test.ts packages/bench/src/providers/openai-compatible.test.ts`
- `pnpm --filter @remnic/core build && pnpm --filter @remnic/bench build && pnpm --filter @remnic/cli build`
- `pnpm exec tsx packages/remnic-cli/src/index.ts bench run --quick ama-bench --runtime-profile real --remnic-config ~/.config/remnic/config.json --system-provider openai --system-base-url http://127.0.0.1:1234/v1 --system-model google/gemma-4-26b-a4b --judge-provider openai --judge-base-url http://127.0.0.1:1234/v1 --judge-model google/gemma-4-26b-a4b`
- confirmed full-mode failure now reports the LM Studio context-window configuration problem explicitly:
  - `OpenAI-compatible completion failed: 400 ... n_keep ... n_ctx ... LM Studio is running this model with a context window that is too small for this benchmark. Increase the model context length in LM Studio and rerun.`

## Notes
- I intentionally did **not** keep the temporary `3000`-char recall cap discovered during local debugging, because that would silently change normal benchmark behavior for users with correctly configured model servers.
- Full non-quick Gemma runs remain throughput-heavy and depend on the local LM Studio model being loaded with a sufficiently large context window.

---

> [!NOTE]
> **Medium Risk**
> Touches process/daemon lifecycle and teardown paths (bench adapter, CLI exit, QMD daemon session pooling), which can cause hangs or premature shutdown if incorrect. No auth or data-model changes, but behavior changes are runtime-sensitive.
> 
> **Overview**
> Hardens local benchmark shutdown to avoid post-run hangs by aborting orchestrator deferred init, clearing QMD maintenance timers/flags, waiting briefly for `deferredReady`, disposing QMD, then closing the LCM engine before removing the temp dir.
> 
> Improves OpenAI-compatible error diagnostics by appending a context-window hint (special-cased for LM Studio `127.0.0.1:1234`) when responses indicate `n_ctx`/context-length issues, with new tests covering both hinted and unchanged error cases.
> 
> Makes CLI completion more reliable by keeping `sync watch` alive via an awaited never-resolving promise and arming a success-exit watchdog that flushes stdout/stderr then forces a clean exit after a short grace period if handles remain open.
> 
> Prevents leaked QMD resources by adding `dispose()` support to `QmdClient` (ref-counted shared daemon sessions) and having `Orchestrator` dispose the current search backend before swapping to `NoopSearchBackend` when collections are missing.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ce902807788bbf844eadd7fa5b9f1b8c772e6d2b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/bench/src/adapters/remnic-adapter.ts
+++ b/packages/bench/src/adapters/remnic-adapter.ts
@@ -5,7 +5,7 @@
 import { mkdir, mkdtemp, rm } from "node:fs/promises";
 import { tmpdir } from "node:os";
 import path from "node:path";
-import { Orchestrator, parseConfig } from "@remnic/core";
+import { Orchestrator, parseConfig, shutdownQmdResources } from "@remnic/core";
 import type {
   BenchJudge,
   BenchMemoryAdapter,
@@ -81,6 +81,15 @@ export const BENCH_ADAPTER_MODE_CONFIG: Record<BenchAdapterMode, Record<string, 
     extractionMinUserTurns: 1000000,
     recallPlannerEnabled: false,
   },
+};
+
+type OrchestratorTeardownView = {
+  abortDeferredInit(): void;
+  deferredReady: Promise<void>;
+  lcmEngine: { close(): void } | null;
+  qmdMaintenanceTimer?: NodeJS.Timeout | null;
+  qmdMaintenancePending?: boolean;
+  qmdMaintenanceInFlight?: boolean;
 };
 
 function cloneBenchConfig(config: Record<string, unknown>): Record<string, unknown> {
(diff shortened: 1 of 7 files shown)


joshuaswarren:
@cursor review

joshuaswarren:
@codex review
@cursor review

joshuaswarren:
@cursor review

PR #599, opened 2026-04-20
feat(xray): shared Recall X-ray renderer (#570 PR 2/7)

joshuaswarren:
## Summary

Slice 2 of the unified Recall X-ray observability surface (#570). Pure renderer only — CLI / HTTP / MCP wiring land in PRs 3-5, which all consume this module (CLAUDE.md rule 22 — single source of truth for formatting).

_Successor to the now-closed #594 (auto-closed when slice 1's base branch was deleted on merge of #578).  The branch has been rebased onto `main` and only contains slice 2's commits._

- New `packages/remnic-core/src/recall-xray-renderer.ts`
  - `renderXray(snapshot, format)` top-level dispatcher. `format ∈ {"json", "text", "markdown"}`.
  - `renderXrayJson`: stable v1 envelope. `{schemaVersion: "1", snapshotFound: false}` for null snapshots, otherwise `{snapshotFound: true, ...snapshot}` as pretty-printed JSON.
  - `renderXrayText`: human-readable plain-text layout — top-level fields, filter ladder (considered/admitted with optional reason), per-result breakdown (served-by, path, score decomposition, admitted-by, rejected-by, graph-path, audit-entry), and a tier-explain block.
  - `renderXrayMarkdown`: GFM equivalent with tables for the summary / filters / tier-explain blocks, H3 sections per result, pipe-escaping so filter names and reasons cannot break the table layout.
  - `parseXrayFormat`: strict input validator that rejects unknown values with a listed-options error (CLAUDE.md rule 51). `undefined` / `null` default to `"text"`.
  - Deterministic 4-decimal score formatting and ISO-8601 timestamp formatting so golden-file tests stay stable across environments.  Out-of-range finite timestamps coerce to `(unknown)` so a corrupted or custom-clock snapshot cannot crash the renderer.

## No behavior change

No existing surface consumes this module yet. Default recall behavior is unchanged from `main`.

## Test plan

- [x] 18 unit tests in `packages/remnic-core/src/recall-xray-renderer.test.ts` — every format, full + minimal + null snapshots, golden text and golden markdown, dispatcher parity, format validator errors, pipe escaping, score formatting, non-finite and out-of-range `capturedAt` fallbacks.
- [x] `pnpm run check-types` clean.

Part of #570 (slice 2 of 7).

---

> [!NOTE]
> **Low Risk**
> Mostly additive rendering + unit tests; the only behavior change is a small refactor of Recall Explain tier-explain text rendering which should remain byte-identical via the new shared helper.
> 
[...]

PR #597, opened 2026-04-20
feat(xray): remnic xray CLI command (#570 PR 3/7)

joshuaswarren:
## Summary

Slice 3 of the unified Recall X-ray observability surface (#570). Adds the `remnic xray \"<query>\"` CLI command that runs a recall with X-ray capture enabled, fetches the resulting `RecallXraySnapshot`, and renders it through the shared renderer from PR 2. CLI / HTTP / MCP all consume the same renderer (CLAUDE.md rule 22).

- New `packages/remnic-core/src/recall-xray-cli.ts` — pure input-validation helpers extracted from the CLI handler so option-parsing paths can be unit-tested without booting an orchestrator.
  - `parseXrayBudgetFlag`: coerces numeric strings to positive integers; rejects zero, negative, fractional, non-finite, and non-numeric values with a listed-options error (CLAUDE.md rules 14 + 51).
  - `parseXrayCliOptions`: validates `<query>`, `--format`, `--budget`, `--namespace`, `--out`. Empty or whitespace-only values collapse to documented defaults; invalid values throw with explicit error messages.

- `packages/remnic-core/src/cli.ts` — new `xray` subcommand alongside `recall-explain`.
  - Flags: `--format {text|markdown|json}`, `--budget <chars>`, `--namespace <ns>`, `--out <path>`.
  - Applies `--budget` by swapping `orchestrator.config.recallBudgetChars` around the recall call in a try/finally so a thrown recall cannot leave the config mutated.
  - Clears any prior x-ray snapshot before the recall so capture failures surface as `null` rather than returning stale data from a prior call.
  - Expands `~` in `--out` via the existing `expandTildePath` helper (CLAUDE.md rule 17).

## No behavior change

No changes to existing recall, recall-explain, or any other CLI surface.

## Stacking

Based on `feat/issue-570-pr2-renderer` (PR #594). Retargets to `main` after PRs #578 and #594 merge.

## Test plan

- [x] 12 unit tests in `packages/remnic-core/src/recall-xray-cli.test.ts` — default format, full option bag, empty/whitespace/non-string query, unknown format, malformed budget (zero, negative, fractional, non-numeric, non-finite), blank namespace/out coercion, case-insensitive format normalization.
- [x] All renderer tests from PR 2 still pass.
- [x] `tsc --noEmit` clean.

Part of #570 (slice 3 of 7).

---

> [!NOTE]
> **Medium Risk**
[...]

Issue #570, opened 2026-04-20
feat(observability): unified recall X-ray surface (tier + audit + MMR in one view)


joshuaswarren:
## Problem

Remnic already has more retrieval observability than almost any competitor: `recall/explain` (graph paths), `tier-explain` (direct-answer tier annotations, #518), `recall-audit` (write/read audit trail), and MMR diversification traces. But they're scattered across three HTTP endpoints, two CLI commands, and different MCP tools, with inconsistent output shapes. Users can't answer 'why did Remnic surface these memories for this prompt?' in one call. Every 2026 landscape article calls out 'when retrieval fails, there's no useful explanation' as a persistent weakness — we should own this and make it a headline feature.

## Goal

One call — `remnic xray <query>` — returns a structured, per-result attribution document that merges:
- Tier that served each result (`direct-answer`, `hybrid`, `graph`, `recent-scan`, `procedural`, `review-context`).
- Score decomposition (vector sim, BM25 rank, importance boost, MMR penalty, tier prior).
- Graph path (if the tier used graph retrieval).
- Audit trail entry ID so users can drill into writes.
- The exact filter/eligibility ladder that either admitted or rejected each candidate.

Package it as 'Memory X-ray' on the site. This is a marketable differentiator competitors don't have.

## Non-goals

- Replacing existing endpoints. They keep working. X-ray is a superset.
- Shipping a TUI. Terminal output is fine; a visualizer is a separate optional package.

## Existing surface area

- `packages/remnic-core/src/recall-audit.ts`
- `packages/remnic-core/src/recall-mmr.ts`
- `packages/remnic-core/src/direct-answer.ts` + `direct-answer-wiring.ts` (tier-explain)
- `packages/remnic-core/src/retrieval.ts` (main orchestration)
- `packages/remnic-core/src/recall-state.ts` (LastRecallSnapshot + tierExplain)
- CLAUDE.md 'Agent Notes: Retrieval Explain Surface (issue #518)' section — the planned unified renderer lives here but isn't built yet.

### PR 1 — `RecallXraySnapshot` schema + in-memory capture

- New `packages/remnic-core/src/recall-xray.ts` defining `RecallXraySnapshot`:
  ```ts
  interface RecallXraySnapshot {
    query: string;
    snapshotId: string;        // UUID per call
    capturedAt: number;
    tierExplain: RecallTierExplain | null;
    results: RecallXrayResult[];
    filters: RecallFilterTrace[];
[...]

PR #601, opened 2026-04-20
feat(xray): HTTP GET /recall/xray route + access-service plumbing (#570 PR 4/7)

joshuaswarren:
## Summary

Slice 4 of the unified Recall X-ray observability surface (#570). Adds the HTTP surface and the shared `EngramAccessService.recallXray` method that CLI / HTTP / MCP will all call into so the three surfaces return identical JSON (CLAUDE.md rule 22).

- `packages/remnic-core/src/access-service.ts` — new `recallXray({query, sessionKey?, namespace?, budget?, authenticatedPrincipal?})` method.
  - Resolves namespace scope through the same `canReadNamespace` path as `recallTierExplain` (CLAUDE.md rule 42) so an unauthorized principal cannot capture an x-ray for a namespace it cannot read.
  - Drops the captured snapshot when the served namespace differs from the requested one so cross-tenant data cannot leak (CLAUDE.md rules 42 + 47).
  - Requires an identity when namespaces are enabled and no namespace is explicitly requested (least-privileged default — rule 48).
  - Rejects empty query and invalid `budget` (not a positive integer) with explicit errors (CLAUDE.md rules 14 + 51).
  - Swaps `orchestrator.config.recallBudgetChars` in a try/finally so a thrown recall cannot leave the config mutated.
  - Always calls `clearLastXraySnapshot()` first so capture failures surface as `{snapshotFound: false}` rather than returning stale data.

- `packages/remnic-core/src/access-http.ts` — new `GET /engram/v1/recall/xray` route.
  - Query params: `q` (required), `session`, `namespace`, `budget`.
  - Sits behind the existing bearer-auth middleware.
  - Invalid `budget` returns 400 with an explicit message; empty `q` returns 400; service-level validation errors surface as 400 `invalid_request`.

## No behavior change

No changes to existing `/recall`, `/recall/explain`, or `/recall/tier-explain` endpoints. The new route is purely additive.

## Test plan

- [x] 11 integration tests in `tests/access-service-recall-xray.test.ts` using lightweight orchestrator stubs — empty-query rejection, absent snapshot → `snapshotFound: false`, present snapshot round-trips, `xrayCapture: true` is forwarded, prior snapshot is cleared before capture, budget override is restored even when recall throws, invalid budgets rejected, namespace read permissions enforced before recall fires, cross-namespace capture drops the snapshot, identity required when namespaces are enabled without a requested namespace.
- [x] `tsc --noEmit` clean for the core package.

## Stacking

Based on `main`. Independent of PRs #597 and #599 — HTTP returns the `RecallXraySnapshot` as-is, it does not invoke the renderer.

Part of #570 (slice 4 of 7).

---

> [!NOTE]
> **Medium Risk**
> Touches recall invocation and namespace/ACL enforcement paths, including new principal and budget overrides; mistakes could leak cross-namespace data or change recall budgeting behavior. Changes are additive but affect core recall plumbing used by multiple access surfaces.
> 
> **Overview**
> Adds a new bearer-authenticated `GET /engram/v1/recall/xray` route that returns a structured Recall X-ray snapshot, with explicit 400s for missing `q` and invalid `budget`.
> 
> Introduces `EngramAccessService.recallXray()` to run recalls with `xrayCapture: true`, validate inputs, enforce namespace read access *before* recall, and drop snapshots if the captured namespace doesn’t exactly match the requested one; includes a per-service mutex to prevent snapshot-slot races.
> 
> Extends `RecallInvocationOptions` with per-call `budgetCharsOverride` and `principalOverride`, and threads these through orchestrator budgeting and principal resolution so x-ray calls avoid mutating shared config; adds integration tests covering validation, ACL behavior, override forwarding, and concurrency safety.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 183896ef1e748db7ac520487552d4ac57a5b2b6f. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/remnic-core/src/access-http.ts
+++ b/packages/remnic-core/src/access-http.ts
@@ -394,6 +394,74 @@ export class EngramAccessHttpServer {
       return;
     }
 
+    // Recall X-ray (issue #570 PR 4): unified per-result attribution
+    // snapshot.  Requires bearer auth (same as every other endpoint
+    // here) and enforces namespace scope before the recall fires
+    // (CLAUDE.md rule 42).  Query comes from the `q` search param so
+    // GET stays cacheable; `namespace` / `session` / `budget` are
+    // optional.
+    if (req.method === "GET" && pathname === "/engram/v1/recall/xray") {
+      const queryParam = parsed.searchParams.get("q");
+      if (!queryParam || queryParam.trim().length === 0) {
+        this.respondJson(res, 400, {
+          error: "missing_query",
+          code: "missing_query",
+          message: "q search parameter is required and must be non-empty",
+        });
+        return;
+      }
+      const sessionParam = parsed.searchParams.get("session");
+      const sessionKey = sessionParam && sessionParam.length > 0
+        ? sessionParam
+        : undefined;
+      const namespaceParam = parsed.searchParams.get("namespace");
+      const namespace = this.resolveNamespace(
+        req,
+        namespaceParam && namespaceParam.length > 0
+          ? namespaceParam
+          : undefined,
+      );
+      const budgetParam = parsed.searchParams.get("budget");
+      // Reject invalid `budget` with 400 rather than silently
+      // defaulting (CLAUDE.md rules 14 + 51).
+      let budget: number | undefined;
+      if (budgetParam !== null && budgetParam !== "") {
+        const parsedBudget = Number(budgetParam);
+        if (
+          !Number.isFinite(parsedBudget)
+          || parsedBudget <= 0
+          || !Number.isInteger(parsedBudget)
+        ) {
+          this.respondJson(res, 400, {
+            error: "invalid_budget",
+            code: "invalid_budget",
+            message:
+              "budget expects a positive integer",
+          });
+          return;
+        }
+        budget = parsedBudget;
+      }
+      try {
+        const payload = await this.service.recallXray({
+          query: queryParam,
+          sessionKey,
+          namespace,
+          budget,
+          authenticatedPrincipal: this.resolveRequestPrincipal(req),
+        });
+        this.respondJson(res, 200, payload);
+      } catch (err) {
+        const message = err instanceof Error ? err.message : String(err);
+        this.respondJson(res, 400, {
+          error: "invalid_request",
+          code: "invalid_request",
+          message,
+        });
+      }
+      return;
+    }
+
     if (req.method === "POST" && pathname === "/engram/v1/observe") {
       const body = await this.readValidatedBody(req, "observe");
       this.ensureWriteRateLimitAvailable();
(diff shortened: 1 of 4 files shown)


joshuaswarren:
Fixed all four review threads in 685273dc:

- **P1 (authenticated principal):** When `authenticatedPrincipal` is present but `sessionKey` isn't, the principal is now threaded in as the recall sessionKey so the orchestrator's internal `resolvePrincipal` evaluates ACLs against the same principal the pre-check authorized.
- **P1 (concurrent budget mutation):** Added a per-service FIFO mutex (`xrayQueue`) that serializes x-ray invocations.  Global config mutation stays until `orchestrator.recall()` accepts a per-call budget; the mutex keeps request isolation intact in the meantime.  New test `recallXray serializes concurrent budget overrides …` deliberately interleaves two calls.
- **P2 (500 vs 400 misclassification):** HTTP route now only 400s validation-class errors (messages prefixed with `recallXray:`); backend faults rethrow to the global handler so they return 500 + get logged.
- **P2 (Cursor — undefined-namespace bypass):** Post-capture guard is now strict; a snapshot whose `namespace` is `undefined` no longer bypasses a requested scope.  New test covers the regression.

14 tests now; all pass.

joshuaswarren:
Addressed Codex P1 + Cursor Low in commit 183896ef. Two architectural fixes:

1. **Codex P1 (principal resolution)**: threading `authenticatedPrincipal` through `recallSessionKey` was wrong — the orchestrator's `resolvePrincipal(sessionKey)` only maps configured raw session keys and otherwise collapses to `"default"`. Added a new `principalOverride` field to `RecallInvocationOptions` and the recall path now resolves its principal via that override when present. In namespace-enabled deployments, ACL checks now use the SAME identity the access-surface pre-check authorized.

2. **Cursor Low (deadlock risk)**: the `orchestrator.config.recallBudgetChars` swap between mutex acquisition and the try/finally could throw and leave `xrayQueue` permanently unresolved, deadlocking future calls. Dropped the config mutation entirely — budget is now threaded via `options.budgetCharsOverride`. The mutex stays only for snapshot-slot ordering.

Tests updated + added:
- Budget threading test asserts shared config is NEVER mutated and the override lands in options
- Principal override test asserts sessionKey is NOT polluted and `principalOverride` carries identity
- Concurrent-recalls test asserts each call sees its own budget and shared config stays unchanged throughout

CLAUDE.md rules 42 (read/write paths resolve through same namespace layer) + 47 (no shared mutable state across async boundaries).

joshuaswarren:
Addressed Cursor Low in commit 21617d8b: trim `sessionKey` once and reuse the same string for the pre-check `resolvePrincipal`, the namespaces-enabled identity guard, and the recall invocation — so whitespace-padded keys can no longer resolve to different principals between the two ACL evaluations. `resolvePrincipal` does not trim internally, which is what exposed the drift.

PR #607, opened 2026-04-20
feat(procedural): safer-by-default thresholds (#567 PR 3/5)

joshuaswarren:
Part of #567 (slice 3 of 5).

## Summary

Raises the floor for procedural mining + recall so the slice-4 default-on flip is safe. This slice only tunes defaults — it does NOT touch `procedural.enabled`.

| Key | Before | After |
| --- | ------ | ----- |
| `procedural.successFloor` | 0.7 | **0.75** |
| `procedural.lookbackDays` | 30 | **14** |
| `procedural.recallMaxProcedures` | 3 | **2** |

`minOccurrences` (3) and `autoPromoteOccurrences` (8) already matched the issue spec; no change there.

Touches:

- `packages/remnic-core/src/config.ts` — new defaults in `parseConfig`. `0` disable path for `minOccurrences` is preserved (CLAUDE.md rule 45).
- Both plugin manifests (`packages/plugin-openclaw/openclaw.plugin.json` + auto-synced `packages/shim-openclaw-engram/openclaw.plugin.json`) now advertise the new defaults in the config schema.
- `packages/remnic-core/src/config.test.ts` — new anti-drift test `parseConfig applies safer-by-default procedural thresholds (issue #567 PR 3/5)`. `procedural.enabled` still asserts `false` here; the flip-to-true test lives in slice 4.
- `docs/procedural-memory.md` — new "Safer-by-default thresholds" table with every default + rationale.

## Test plan

- [x] `node --test --import tsx packages/remnic-core/src/config.test.ts` → 46/46 pass (45 existing + 1 new).
- [x] `pnpm --filter @remnic/core run check-types` → clean.
- [x] Shim manifest re-synced via `packages/shim-openclaw-engram/scripts/sync-openclaw-plugin.mjs` so defaults stay consistent across both.

No behavioral change until `procedural.enabled=true` — current default is still `false`.

---

> [!NOTE]
> **Low Risk**
> Low risk: small behavioral change only when `procedural.recallMaxProcedures` is missing or invalid, reducing injected procedure previews from 3 to 2 to match the canonical default.
> 
> **Overview**
> Ensures procedural recall uses the **safer-by-default** cap when `procedural.recallMaxProcedures` is absent or non-finite by changing the fallback in `buildProcedureRecallSection` from `3` to `2` (matching `config.ts`).
> 
> Updates the procedural memory docs to correctly describe the default as being *lowered* from `3` to `2`.
> 
[...]

Issue #567, opened 2026-04-20
feat(procedural): validate lift on bench + flip procedural.enabled default to true


joshuaswarren:
## Problem

Procedural memory shipped via #519 but ships **disabled by default** (`procedural.enabled: false`). 0% of users benefit today. The 2026 academic wave (LEGOMem, Dynamic Procedural Memory Framework, A-Mem) and competitors (Hindsight biomimetic, ByteRover skill memory) are making procedural / skill memory a default expectation. We need to (a) prove it helps with a bench number, (b) flip the default to `true` with safe bounded defaults, (c) communicate the change clearly.

## Goal

Ship procedural memory ON by default, backed by a published bench number showing lift vs. off.

## Non-goals

- Rewriting the procedural-mining algorithm. `@remnic/core` already has one.
- Changing the default thresholds until the bench run justifies them.

## Existing surface area

- Module: `packages/remnic-core/src/procedural/` (if present) — verify current shape first.
- Bench scenario: `packages/bench/src/benchmarks/remnic/procedural-recall/`
- Config: `procedural.*` in `config.ts`, `packages/plugin-openclaw/openclaw.plugin.json`, `packages/shim-openclaw-engram/openclaw.plugin.json`.
- Docs: `docs/procedural-memory.md`

### PR 1 — Bench ablation harness

- Add `packages/bench/src/benchmarks/remnic/procedural-recall/ablation.ts` that runs the same fixture twice (procedural.enabled=false vs. true) and emits a diff artifact `{onScore, offScore, lift, confidenceInterval}`.
- CLI: `remnic bench procedural-ablation --fixture <path> --out <path>`.
- Tests: stub LLM adapter so the harness runs deterministically in CI.

### PR 2 — Procedural bench on real fixtures

- Extend the existing `procedural-recall/` fixtures with ≥20 procedure-recall scenarios covering: exact re-run, parameter variation, decomposition, distractor rejection.
- Run the ablation from PR 1; commit a baseline `procedural-recall-baseline.json`.
- Write-up in `docs/benchmarks/procedural-recall.md` with the numbers.

### PR 3 — Safer-by-default config sweep

- If PR 2's lift is ≥ 3 points with zero regression on LongMemEval-S / LoCoMo smoke, proceed. Otherwise raise thresholds until regression gone.
[...]

Issue #569, opened 2026-04-20
feat(coding-agent): project/branch-scoped namespaces + review-context recall


joshuaswarren:
## Problem

ByteRover owns the 'coding-agent memory' narrative in 2026 by offering version-control semantics for memory (branches, commits, merges of context). Supermemory has the widest MCP coverage for Claude Code / OpenCode. Remnic's existing Claude Code / Codex CLI / Cursor adapters inject a *flat* memory scope — whatever's in the default namespace for the user — so memories from project A leak into a session working on project B, and memories from a feature branch can contaminate recall on `main`.

Remnic already has a namespace subsystem (`packages/remnic-core/src/namespaces/`). It just isn't wired to git context at session-start.

## Goal

When an agent opens a session inside a git repo, Remnic:
1. Auto-scopes memory to the project (git remote origin URL or root path hash).
2. Overlays an optional per-branch namespace for branch-specific experiments.
3. On review-style prompts ('review this PR', 'what changed in this diff'), ranks recall by files actually touched in the diff.

This gives Claude Code / Codex CLI users a reason to choose Remnic over ByteRover: same version-control mental model, but local-first and multi-host.

## Non-goals

- A UI for manually branching memory. Just implicit scoping based on git state.
- Hosting / sharing memory across team members. That's the Spaces (M6) territory; separate issue if needed.
- GitHub PR API integration. This feature only needs a local diff.

## Existing surface area

- `packages/remnic-core/src/namespaces/principal.ts` — namespace resolution
- `packages/remnic-core/src/namespaces/search.ts` — namespace-scoped search
- `packages/remnic-core/src/connectors/codex/` — connector pattern
- Claude Code adapter: check for `packages/connector-claude-code/` or equivalent
- `Orchestrator` session-start hook: `before_agent_start`
- Recall pipeline: `packages/remnic-core/src/retrieval.ts`

### PR 1 — `GitContextResolver` module (pure, no wiring)

- New `packages/remnic-core/src/coding/git-context.ts`.
- `resolveGitContext(cwd): Promise<{ projectId, branch, rootPath, defaultBranch } | null>`.
- `projectId` = stable hash of `origin` URL (fallback to root path when origin absent).
- Pure function: no orchestrator, no config side-effects.
- Tests: mocked `git` invocations, cwd outside a repo returns `null`.
- CLAUDE.md rule 17: expand `~` in root path.

### PR 2 — Project-scoped namespace overlay

[...]

PR #609, opened 2026-04-20
feat(procedural): flip procedural.enabled default to true (#567 PR 4/5)

joshuaswarren:
Part of #567 (slice 4 of 5).

**Base branch:** `feat/issue-567-pr3-safer-defaults` (slice 3). GitHub will retarget to `main` after #607 merges.

## Summary

Flips the master gate: procedural memory now ships **enabled by default**.

- `parseConfig` defaults `procedural.enabled` to `true` when the key is omitted. Explicit `false`, `"false"`, `"0"`, `"no"`, `"off"` continue to opt out (CLAUDE.md rule 36). Unrecognized string values fall back to the default (true).
- Both plugin manifests advertise `default: true` with an explicit opt-out note in the description.
- Exhaustive `config.test.ts` case covers every path: omitted → `true`, each falsy-string variant → `false`, explicit boolean → honored, unrecognized string → falls back.
- Docs sync across README, `docs/procedural-memory.md`, `docs/config-reference.md`, `docs/getting-started.md`, `CLAUDE.md`, and `CHANGELOG.md`.

## CLAUDE.md invariants honored

- **Rule 30** (config gate with working disable) — opt-out is an explicit `false` (and falsy-string variants), never silently overridden.
- **Rule 36** (string "false" is truthy) — verified by test: `"false"` / `"0"` / `"no"` / `"off"` all coerce to off.
- **Rule 51** (reject invalid input) — unrecognized string values fall back to the documented default, same as every other `coerceBool`-gated field in the config.
- **Rule 55** (documented behavior wired + tested) — every changed default has a test assertion and a doc row.
- **À-la-carte packaging** — this is a `@remnic/core` config change only; no bench/plugin dependency is added to the base install.

## Test plan

- [x] `node --test --import tsx packages/remnic-core/src/config.test.ts` → 47/47 pass (46 existing + 1 new).
- [x] `node --test --import tsx tests/procedure-miner.test.ts tests/procedure-recall.test.ts` → 4/4 pass (existing explicit configs unaffected).
- [x] `pnpm --filter @remnic/core run check-types` → clean.
- [x] Manual verification: `remnic init` with no explicit `procedural` key produces `procedural.enabled === true` (asserted by the new config test which exercises the same `parseConfig` code path).

## Merge order

Depends on #607 (slice 3). Merge order should be **slice 3 → slice 4**. If slice 3 merges first, GitHub will retarget this PR to `main` automatically. The auto-sync shim manifest run shows no unexpected diff relative to the plugin-openclaw source.

---

> [!NOTE]
> **Medium Risk**
> Changes a default feature gate (`procedural.enabled`) from off to on, which will start writing procedure memories and injecting procedure recall for users who haven’t explicitly configured it. Also tightens config parsing/merging behavior, so mis-shaped configs that previously “worked” may now throw.
> 
> **Overview**
> **Procedural memory now ships enabled by default**: `parseConfig` treats an omitted `procedural.enabled` as `true`, and the default recall pipeline now enables the `procedure-recall` section unless explicitly opted out.
> 
> Config handling is hardened: the `conservative` preset explicitly keeps `procedural.enabled` off, preset/user config merging deep-merges the `procedural` block to avoid accidentally re-enabling, and non-object `procedural` shapes or unparseable `procedural.enabled` values now throw instead of failing open.
> 
> Plugin manifests and docs are updated to reflect the new defaults and safer thresholds (`successFloor=0.75`, `lookbackDays=14`, `recallMaxProcedures=2`), with expanded unit tests locking in default/opt-out/invalid-input behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit d9a04dddb58e6d677aa813865b61343e3e729bf4. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,19 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Changed
+
+- **Procedural memory is now enabled by default** (issue #567 PR 4/5;
+  previously shipped disabled). Fresh installs and any config that omits
+  `procedural.enabled` now get procedure extraction, task-initiation
+  recall injection, and trajectory mining using the safer-by-default
+  thresholds from #567 PR 3 (`successFloor=0.75`, `lookbackDays=14`,
+  `recallMaxProcedures=2`). Operators who want to stay opt-out should
+  set `procedural.enabled: false` explicitly. See
+  [`docs/procedural-memory.md`](docs/procedural-memory.md) and the
+  baseline lift artifact in
+  [`packages/bench/baselines/procedural-recall-baseline.json`](packages/bench/baselines/procedural-recall-baseline.json).
+
 ## [v9.3.103] — 2026-04-20
 
 ### Added
--- a/CLAUDE.md
+++ b/CLAUDE.md
@@ -26,7 +26,7 @@
 
 ### Procedural memory (issue #519)
 
-Ships **disabled** until plugin config sets **`procedural.enabled`: `true`** (nested `procedural` object). Operators and agents should read `docs/procedural-memory.md` and the README **Configuration** table before enabling.
+Ships **enabled by default** since issue #567 PR 4/5 (previously default-off). Operators who want to opt out must set **`procedural.enabled`: `false`** (nested `procedural` object). Agents should read `docs/procedural-memory.md` and the README **Configuration** table for the full threshold defaults and the opt-out path.
 
 ### Before every commit, verify:
 - `git diff --cached` contains NO personal information
--- a/README.md
+++ b/README.md
@@ -566,7 +566,7 @@ Run Engram as a standalone HTTP server that multiple agent harnesses share. Isol
 - **Extraction Judge** — LLM-as-judge post-extraction filter that evaluates fact durability before write. Has shadow mode for calibration. Opt-in via `extractionJudgeEnabled`. (issue #376)
 - **Semantic Chunking** — Smoothing-based topic boundary detection using sentence embeddings and cosine similarity, as an alternative to recursive chunking. Opt-in via `semanticChunkingEnabled`. (issue #368)
 - **OAI-mem-citation Blocks** — Recall responses emit `<oai-mem-citation>` blocks matching the Codex citation format for memory attribution and usage tracking. Opt-in via `citationsEnabled`. (issue #379)
-- **Procedural memory** — Stores repeatable **how-to** memories as `category: procedure` markdown under `procedures/`, mines candidates from causal trajectories, and can inject a **Relevant procedures** section on task-initiation prompts. **Off by default**; set `procedural.enabled` to `true` in plugin config. See [Procedural memory](docs/procedural-memory.md). (issue #519)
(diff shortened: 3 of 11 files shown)


joshuaswarren:
Addressed the four review threads in 2e9f3c1:

1. **Codex P1 — fail-open on invalid enabled values** (PRRT...CWY): `parseConfig` now throws when `procedural.enabled` is set but unrecognized (typo strings like `"fales"`, numbers like `0` or `1`, `null`, empty/whitespace strings). Only an omitted key falls back to the default-on behavior. Test expanded to cover each rejection class.
2. **Codex P2 + Cursor (Medium) — root manifest drift** (PRRT...CWc / DmA): re-ran `scripts/sync-openclaw-plugin.mjs` so the root `openclaw.plugin.json` matches the packages source. The "root and package manifests stay byte-identical" test is now green on this branch.
3. **Cursor (Low) — duplicate scenario** (PRRT...DmH): the `bareConfig` branch now passes `procedural: {}` explicitly, distinct from the `omitted` case.

All four threads resolved.

joshuaswarren:
Addressed the two follow-up threads in fa59be5:

1. **Codex P2 — procedure-recall pipeline stayed disabled when key omitted**: `buildDefaultRecallPipeline` now mirrors `parseConfig`'s default-on semantics (undefined → enabled; coerceBool false → disabled). `parseConfig` still rejects invalid values loudly at parse time, so this branch only ever sees well-formed input. New test `buildDefaultRecallPipeline enables procedure-recall when procedural default-on` locks in the behavior.
2. **Cursor (Medium) — conservative preset silently enabled procedural**: the `conservative` memoryOsPreset now pins `procedural: { enabled: false }` so the preset stays restrictive after the default flip. User overrides still win (preset ← baseCfg), so `conservative + procedural:{enabled:true}` restores the feature. New test `conservative memoryOsPreset keeps procedural.enabled off` covers both paths.

Both threads resolved.

joshuaswarren:
Addressed the Codex P1 follow-up in d1ab95d:

**Codex P1 — partial procedural override wiped conservative opt-out**: the shallow `{...preset, ...baseCfg}` merge meant a user-provided `procedural: { minOccurrences: 5 }` (no `enabled` key) would replace the preset's `procedural: { enabled: false }` wholesale. `parseConfig` now deep-merges the `procedural` block when both sides supply one — preset keys fill gaps, baseCfg keys override specific fields. Extended test confirms `conservative + procedural:{minOccurrences:5}` still yields `enabled:false` with `minOccurrences:5` honored.

The Cursor Medium on `codingContext null clearing` (claude-code-plugin) and the Codex P2 on `resolveMemoryDir` fallback are both about code that doesn't exist on this branch (slice 4 is core config + docs only; neither claude-code-plugin connector nor CLI procedural stats live here). Marking both outdated/resolved as out-of-scope for this PR — slice 5 (#611) shipped the procedural-stats CLI with the proper memoryDir resolution; the codingContext clearing is issue #569 territory.

joshuaswarren:
Cursor Medium addressed in 4fee2d0: the `procedure-recall` pipeline gate now uses `coerced === true` instead of `coerced !== false`, so invalid values fall closed. `parseConfig` still rejects invalid `procedural.enabled` upstream, but the defense-in-depth fix matches CLAUDE.md rule 48 (least-privileged defaults) and the fail-closed convention everywhere else in config.

joshuaswarren:
Codex P2 addressed in d9a04dd: `parseConfig` now throws on non-object `procedural` shapes (`false`, `null`, numbers, strings, arrays) instead of silently normalizing to `{}` and enabling the feature. Use `procedural: { enabled: false }` for opt-out. New test covers every rejection class.

Issue #568, opened 2026-04-20
feat(import): memory importers for ChatGPT, Claude, Gemini, Mem0


joshuaswarren:
## Problem

Google Gemini shipped a feature in March 2026 that imports chat history + memory from ChatGPT and Claude. This is normalizing the idea that AI memory is portable. Remnic's positioning — one shared memory across every agent — is weakened if switching *into* Remnic requires retyping everything. Users who invested in ChatGPT/Claude/Gemini/Mem0 should be able to bring their accumulated memory into Remnic in under 5 minutes.

We already have a WeClone importer pattern (`@remnic/import-weclone`). Extend the same pattern.

## Goal

Ship CLI importers for the four dominant memory surfaces: ChatGPT, Claude, Gemini, and Mem0. Each maps the source's native export format to Remnic memories with source attribution.

## Non-goals

- Two-way sync. This is one-shot import only.
- Real-time listeners / webhook connectors. Separate issue.
- Importing OpenAI conversation embeddings. Only user-facing memory content.

## Existing surface area

- Template to copy: `packages/import-weclone/` (the WeClone importer, landed via #515).
- Orchestrator entry point: `Orchestrator.bulkImport(memories, { sourceLabel, dryRun })`.
- Source attribution: `packages/remnic-core/src/source-attribution.ts`.
- Optional-package loader pattern: `packages/remnic-cli/src/optional-bench.ts` (CLAUDE.md à-la-carte rule).
- CLI dispatcher: `packages/remnic-cli/src/index.ts`, resolves sub-packages via computed dynamic import.

## Architecture contract (MUST follow)

- Each importer is its own optional workspace package: `@remnic/import-chatgpt`, `@remnic/import-claude`, `@remnic/import-gemini`, `@remnic/import-mem0`.
- All four are loaded via computed dynamic import with a clean install-hint on miss (CLAUDE.md à-la-carte rule, rule 57).
- They share a small abstract base in `@remnic/core`: `packages/remnic-core/src/importers/base.ts` exposing `ImporterAdapter` interface (`parse`, `transform`, `writeTo(orchestrator)`).
- Every imported memory MUST carry `sourceLabel`, `importedFromPath`, `importedAt`, and a provenance block.

### PR 1 — Shared importer base + `ImporterAdapter` interface

- Add `packages/remnic-core/src/importers/base.ts` with `ImporterAdapter` + `RunImportOptions { dryRun, rateLimit, batchSize, onProgress }`.
[...]

PR #610, opened 2026-04-20
feat(import): --all-from-bundle auto-detect (#568 PR 7/7)

joshuaswarren:
Part of #568 (slice 7 of 7).

## Summary
- New \`remnic import --all-from-bundle <dir>\` mode walks an unzipped export bundle, detects supported files by name, and runs each matching adapter in sequence.
- Detection walks up to four directory levels (enough to reach Takeout's \`Takeout/Gemini/My Activity.json\`).
- \`conversations.json\` is disambiguated between ChatGPT and Claude by peeking at the first conversation's shape (\`mapping\` → chatgpt, \`chat_messages\` → claude).
- Bundle mode is mutually exclusive with \`--adapter\` / \`--file\` — combining them throws a user-facing error (CLAUDE.md rule 51).
- Per-entry failures surface to stderr without aborting the walk.

## Filename map

| Filename                                  | Adapter   |
|-------------------------------------------|-----------|
| \`memory.json\`                             | chatgpt   |
| \`projects.json\`                           | claude    |
| \`conversations.json\`                      | chatgpt / claude (by content) |
| \`My Activity.json\` / \`MyActivity.json\`    | gemini    |
| \`mem0.json\` / \`mem0-export.json\`          | mem0      |

## Test plan
- [x] \`pnpm run check-types\` clean
- [x] \`packages/remnic-cli\` tests: 36 pass (new: 9 import-bundle-detect tests + 7 parseImportBundleArgs + 3 runBundleImportCommand integration tests)
- [x] Bundle detection is filesystem-injected for tests so no real file-system is touched

---

> [!NOTE]
> **Medium Risk**
> Adds a new multi-import execution path that walks user-supplied directories and changes `cmdImport` return shape; mistakes could affect CLI behavior, resource cleanup, or filesystem safety (mitigated by explicit symlink rejection and tests).
> 
> **Overview**
> Adds a new `remnic import --all-from-bundle <dir>` mode that scans an unzipped export directory, auto-detects supported export files, and runs the appropriate importer adapters sequentially while continuing past per-entry failures and reporting an aggregate failure count.
> 
> Introduces `detectBundleEntries` with stable ordering, `conversations.json` content-based disambiguation (ChatGPT vs Claude), and explicit **symlink rejection** (including symlinked bundle roots) to prevent importing from unintended filesystem locations. Updates `cmdImport` to route to bundle mode, memoize the write target across multiple per-entry runs, and expands the test suite (new detector tests + bundle-args/runner tests) and CLI test script.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2ac55b3095ad331fe5b6cc7086fee65d0d6abc10. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/remnic-cli/package.json
+++ b/packages/remnic-cli/package.json
@@ -27,7 +27,7 @@
     "prebuild": "node ../../scripts/ensure-cli-bench-build-deps.mjs",
     "build": "tsup --config tsup.config.ts && node scripts/copy-bench-assets.mjs",
     "check-types": "tsc --noEmit",
-    "test": "tsx --test src/import-dispatch.test.ts src/optional-importer.test.ts src/bench-args.test.ts",
+    "test": "tsx --test src/import-dispatch.test.ts src/import-bundle-detect.test.ts src/optional-importer.test.ts src/bench-args.test.ts",
     "prepublishOnly": "npm run build"
   },
   "dependencies": {
--- a/packages/remnic-cli/src/import-bundle-detect.test.ts
+++ b/packages/remnic-cli/src/import-bundle-detect.test.ts
@@ -1,0 +1,149 @@
+import assert from "node:assert/strict";
+import { describe, it } from "node:test";
+
+import {
+  detectBundleEntries,
+  type BundleDetectOptions,
+} from "./import-bundle-detect.js";
+
+/**
+ * Build a fake filesystem map that the detect helpers can query through
+ * injected `readdirImpl` / `readFileImpl` / `isDirectoryImpl` hooks. Keys
+ * ending in `/` are directories; every other path is a file whose contents
+ * live in `files[path]` (or default to `{}`).
+ */
+function makeFs(files: Record<string, string>, dirs: string[]): BundleDetectOptions {
+  const dirSet = new Set(dirs);
+  return {
+    readdirImpl: (dir: string) => {
+      const normalized = dir.endsWith("/") ? dir : dir + "/";
+      const all = new Set<string>();
+      for (const key of [...Object.keys(files), ...dirs]) {
+        if (key === dir) continue;
+        if (key.startsWith(normalized)) {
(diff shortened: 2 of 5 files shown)


joshuaswarren:
Addressed the four unresolved threads:

- **P1 (partial bundle failures)** — `cmdImport` now sets `process.exitCode = 1` and emits a stderr summary when any detected entry fails. The continue-on-error behavior is preserved (users still see every failure in one pass), but shell pipelines like `remnic import --all-from-bundle ... && ...` now short-circuit correctly.
- **P2 (symlink directory escape)** — `defaultIsDirectory` uses `lstatSync` and returns `false` for symlinks, so the walker refuses to recurse through symlinked directories. Bundle-root symlinks are also rejected with a user-facing error.
- **P2 (symlink file escape)** — `defaultIsRegularFile` uses `lstatSync` and returns `false` for symlinks; `collectCandidatePaths` now requires `isRegularFile(p)` before adding a candidate, so `classifyFile` never sees symlinked `memory.json` / `projects.json` / etc. paths.
- **Low (targetFactory resource leak)** — `getWriteTarget` memoizes the first materialization via `materializedTarget` + `materializePromise`. The single-construct invariant is now enforced at the io boundary rather than implicitly at the call site, so a non-singleton factory no longer leaks N-1 resources per bundle entry.

Added regression tests: `rejects symlinked files even when their name matches a known shape` and `rejects symlinked directories and does not recurse into them`. `BundleDetectOptions` now exposes `isRegularFileImpl` for test injection, and when callers inject `readdirImpl` / `isDirectoryImpl` the walker derives `isRegularFile` from the injected traversal layer (so tests don't accidentally probe the real filesystem).

All 43 `@remnic/cli` tests pass (`npx tsx --test src/*.test.ts`).

PR #631, opened 2026-04-21
feat(bench): buffer-surprise-trigger A/B benchmark (issue #563 PR 4/4)

### LINKED ISSUE #563: feat(buffer): surprise-gated flush trigger (D-MEM)
## Problem

`packages/remnic-core/src/buffer.ts` flushes based on turn count, char count, and `scanSignals` heuristic signal-level detection. It has no notion of *novelty* — a turn that contradicts existing memory or introduces a preference shift looks the same as a routine "keep going" turn until signal patterns fire.

## Paper

**D-MEM — Dopamine-Gated Routing**, Mar 2026, https://arxiv.org/abs/2603.14597

Uses a reward-prediction-error (RPE) analog to gate whether an incoming turn triggers the expensive consolidation path. Routine / low-RPE inputs hit an O(1) buffer; contradictions and preference shifts hit the O(N) evolution pipeline. Reports >80% token reduction vs A-MEM-style append-and-evolve while preserving recall quality.

Remnic is less exposed than A-MEM because extraction is already async/periodic — but our flush decision is still heuristic, not novelty-aware. A cheap "surprise" score (embedding distance between the new turn and top-K recent memories) is a good first approximation of RPE for our pipeline.

## Goal

Add surprise-above-threshold as a new buffer flush trigger. Feature-flag it, measure against LongMemEval, flip default on win.

### PR 1 — Surprise helper
- New pure function in `buffer.ts` (or a sibling): `computeSurprise(turn, recentMemories, { embedFn, k })` → number in [0, 1].
- Unit tests with mock embedFn on canned corpora.
- Not wired into the buffer yet.

### PR 2 — Buffer trigger (flag-gated)
- Extend `evaluate()` in `buffer.ts` to check surprise when signal-level is `low` and turn-count hasn't hit threshold.
- Gated on config flag `surpriseFlushEnabled` (default `false`) with tunable threshold.
- Tests: synthetic buffer + high-surprise turn → flush fires even below count threshold.

### PR 3 — Telemetry
- Log surprise scores at every flush decision (not just when it fires) so we can tune the threshold from real traffic.
- Surface distribution in `remnic stats`.

### PR 4 — Benchmark + default flip
- LongMemEval A/B with surprise-flush on vs off.
- If measurable lift (or neutral + lower extraction latency), flip default and document.

## Out of scope

- An actual RPE neural model — the embedding-distance proxy is the whole point of this issue.
- Replacing signal-based triggers; surprise is additive.

joshuaswarren:
## Summary

Hermetic A/B benchmark that replays synthetic conversations with annotated topic shifts through `SmartBuffer` with the surprise gate both off (control) and on (candidate), reporting per-task and aggregate F1 / flush-count deltas. Stacked on #627 (PR 3/4).

## Why the default is NOT flipped in this PR

The benchmark uses a deterministic word-bucket embedder so results are stable across machines with no network/LLM/QMD dependency. That determinism has a cost: the hash embedder flags nearly every English turn as "surprising" relative to its neighbors because sparse content words rarely repeat verbatim. Absolute F1 numbers are low and noisy, so the benchmark is best read as a **regression gate** — "surprise-on keeps the additive-only invariant and does not regress below the control" — rather than a source of truth for flipping the production default.

Flipping `bufferSurpriseTriggerEnabled: true` by default requires a real-embedder benchmark run, which is tracked as a follow-up to this PR. The flag stays off.

## What the benchmark measures

Per task:
- `candidate_topic_shift_f1` / `control_topic_shift_f1` (±1-turn tolerance window; token-set F1)
- `topic_shift_f1_delta` — primary movement metric
- `candidate_flush_count` / `control_flush_count`
- `candidate_mean_turns_between_flushes` / control equivalent

## Test plan

- [x] `npx tsx --test packages/bench/src/benchmarks/remnic/buffer-surprise-trigger/runner.test.ts` — 10/10
- [x] Registry wires the benchmark into the public listing
- [x] Additive-only invariant verified by the `full`-mode runner test: on every fixture case, `candidate_flush_count >= control_flush_count`
- [x] `topicShiftF1` unit tests cover perfect match, ±1 tolerance, out-of-tolerance rejection, empty-vs-empty, partial match, and duplicate-prediction assignment

Closes #563 once merged (after #618 + #627).

---

> [!NOTE]
> **Low Risk**
> Primarily adds new benchmark code and test coverage, plus default-disabled config schema fields; no production behavior changes unless the new limiter is explicitly enabled.
> 
> **Overview**
> Introduces the new `buffer-surprise-trigger` benchmark: a deterministic A/B runner that replays short synthetic conversations through `SmartBuffer` with `bufferSurpriseTriggerEnabled` **off vs on**, then reports per-case and aggregate metrics including flush counts and topic-shift F1 (with ±1-turn tolerance) plus a `topic_shift_f1_delta` comparison signal.
> 
> Adds a curated fixture corpus + smoke subset, a comprehensive runner test suite (including an additive-only invariant: candidate flushes never fewer than control), and registers the benchmark in `packages/bench` so it appears in `listBenchmarks`/`getBenchmark` (with updated registry tests).
> 
> Separately extends `openclaw.plugin.json` config schema with new, **default-off** knobs for a cross-namespace recall budget limiter (`recallCrossNamespaceBudget*`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 2a515c06d4b4427a7efaa097dc7da0159b09f438. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/bench/src/benchmarks/remnic/buffer-surprise-trigger/fixture.ts
+++ b/packages/bench/src/benchmarks/remnic/buffer-surprise-trigger/fixture.ts
@@ -80,7 +80,7 @@ export const BUFFER_SURPRISE_TRIGGER_FIXTURE: BufferSurpriseTriggerCase[] = [
       "Should I side-dress with compost or use a liquid fertilizer?",
       "My soil pH tested at 6.8 last month, which seems okay.",
       // Topic shift 2: gardening → finance
-      "Actually I wanted to ask about Treasury I-bonds.",
+      "On another note, I wanted to ask about Treasury I-bonds.",
       "How does the inflation adjustment compound over time?",
       "What's the annual purchase limit per person?",
       "Are they a better deal than TIPS right now?",
@@ -108,7 +108,7 @@ export const BUFFER_SURPRISE_TRIGGER_FIXTURE: BufferSurpriseTriggerCase[] = [
       "Token bucket vs leaky bucket — which fits bursty traffic better?",
       "I was leaning toward a sliding-window counter with Redis.",
       // Shift 1: engineering → cooking
-      "Unrelated — I want to try making sourdough this weekend.",
+      "Unrelated — thinking about making sourdough this weekend.",
       "My starter has been in the fridge for two weeks, is it still alive?",
       "Should I refresh it twice before baking?",
       // Shift 2: cooking → history
--- a/packages/bench/src/benchmarks/remnic/buffer-surprise-trigger/fixture.ts
+++ b/packages/bench/src/benchmarks/remnic/buffer-surprise-trigger/fixture.ts
@@ -1,0 +1,129 @@
+/**
+ * Fixture for the `buffer-surprise-trigger` benchmark (issue #563 PR 4).
+ *
+ * Each case represents a short synthetic conversation with 2–3 annotated
+ * topic shifts. A "topic shift" is the turn where the conversation moves
+ * to a semantically distant subject — that is the turn we expect the
+ * D-MEM surprise gate to flush on when it's working.
+ *
+ * Conversations are intentionally short (12–15 turns) and written in
+ * plain English so the deterministic hash embedder used by the runner
+ * produces meaningful cosine similarities across topic boundaries
+ * without requiring a real embedding model. The runner does NOT call any
+ * LLM, network, or QMD instance — everything is in-memory and
+ * reproducible by seed.
+ *
+ * Each `topicShiftTurnIndices` is 0-indexed into the `turns` array. The
+ * first turn (index 0) is never a topic shift by construction — there is
+ * no prior context to shift away from.
+ */
+
+export interface BufferSurpriseTriggerCase {
+  id: string;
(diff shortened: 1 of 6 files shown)


joshuaswarren:
@cursor review

joshuaswarren:
@cursor review

joshuaswarren:
Reverted the manifest sync commit. The recallCrossNamespaceBudget* fields belong to PR #638 which added both the manifest entries and the config.ts/types.ts implementation. This PR branch doesn't have that code, so the manifests should not include those fields. The byte-identity test failure is a pre-existing main-branch divergence that will resolve when this PR is rebased on main.

joshuaswarren:
@cursor review

joshuaswarren:
@cursor review

joshuaswarren:
@cursor review

PR #721, opened 2026-04-26
feat(console): structured engine-state aggregator (#688 PR 1/3)

joshuaswarren:
## Summary

Issue #688 PR 1/3 — ships **only the data layer + a CLI flag** for the upcoming operator console. No TUI, no HTTP, no MCP wiring, no SSE — those land in PR 2/3 and PR 3/3.

- New `console/state.ts` module exporting `gatherConsoleState()` and the `ConsoleStateSnapshot` interface covering `bufferState`, `extractionQueue`, `dedupRecent`, `maintenanceLedgerTail`, `qmdProbe`, `daemon`, and `capturedAt`.
- Each subsystem read is independently wrapped in try/catch so one failure cannot crash the whole snapshot — failed sections fall back to empty values and append a string to `errors`.
- Output is JSON-serializable so future HTTP / MCP / TUI surfaces can pipe it directly through `JSON.stringify`.
- New CLI subcommand `remnic engram console --state-only` prints the snapshot as JSON and exits. Without `--state-only` it prints a "TUI not yet available" hint and exits non-zero.
- Aggregator uses a permissive `ConsoleStateOrchestratorLike` duck type so future slices can wire in live extraction-queue / dedup / daemon data via optional accessors without churning the contract.

## Tests

`packages/remnic-core/src/console/state.test.ts` — 7 cases covering JSON-serializability, single-subsystem failure isolation, QMD failure isolation, missing-accessor placeholders, optional-accessor population, ledger-tail bounding with malformed-row tolerance, and missing-ledger graceful return.

## Test plan

- [x] tsc clean from packages/remnic-core
- [x] state.test.ts — 7/7 pass
- [x] npm run build succeeds
- [x] preflight:quick — 46 pre-existing failures verified identical on main; zero new failures introduced

---

> [!NOTE]
> **Medium Risk**
> Introduces a new CLI subcommand and a new state-aggregation layer that reads runtime state and streams JSONL ledger files from disk; while read-only, the new filesystem/JSON parsing paths could affect operator workflows and performance on large ledgers.
> 
> **Overview**
> Adds an initial operator-console data layer via new `console/state.ts`, exposing `gatherConsoleState()` to collect a **JSON-serializable** engine snapshot (buffer stats, optional extraction/dedup tails, maintenance-ledger tail, QMD probe info, and daemon metadata) while isolating subsystem failures into an `errors` array.
> 
[...]

Issue #688, opened 2026-04-25
Live engine introspection (iii Console-style runtime debugger)


joshuaswarren:
## Summary

Add a **live engine introspection surface** — a runtime debugger that shows the current state of Remnic's pipelines (buffer, extraction queue, dedup state, consolidation candidates, maintenance ledger) as they happen. Distinct from Recall X-ray (retrieval inspection); this inspects the engine itself.

## Existing infrastructure to extend

- **Operator-toolkit** (`packages/remnic-core/src/operator-toolkit.ts`) is the existing aggregator for `remnic doctor` — checks workspace hygiene, plugin entry, native-knowledge state, namespaces, eval harness, graph health, session integrity, and more. Already pulls from many engine subsystems.
- **Maintenance ledger** (`maintenance/observation-ledger-utils.ts`) — last-N events available.
- **QMD debug** — `qmd.ts:1011` exposes `debugStatus()` returning probe state.
- **Recall X-ray** (#570) — already inspects retrieval; this issue is the engine-side complement.
- **Admin console** (`admin-console/public/`) already has a Maintenance pane (see #691). The engine-introspection HTTP/MCP surface should feed both the new `remnic console` CLI **and** an extended Maintenance pane.

The console aggregator should reuse `operator-toolkit` patterns rather than re-implementing engine state queries.

**SSE for live updates is not in `access-http.ts` today** (no `text/event-stream` matches) — adding it is part of the work; the same SSE channel will serve #691's brain-graph live updates.

## Background and motivation

[agentmemory](https://github.com/rohitg00/agentmemory) ships an "iii Console" for trace-level engine inspection — operators can see exactly what the engine is doing in real time, not just what it returned for a recall query.

Remnic operators today read `~/.openclaw/logs/gateway.log` with grep filters. We have rich telemetry inside the engine (buffer state, extraction-queue health, dedup decisions, maintenance ledger), but no first-class operator surface for live inspection.

## Goals

- One CLI command (`remnic console`) opening a live, scrollable view of:
  - Buffer state (turns since last flush, bytes, surprise score).
  - Extraction queue (depth, recent verdicts, defer reasons).
  - Recent dedup decisions (admitted, suppressed, why).
  - Consolidation candidates (in flight, gates passed/failed).
  - Maintenance ledger (last 50 events).
  - QMD probe state, daemon health.
- Same data via structured HTTP/MCP for tooling.
- Pull-only by default; push (SSE) wired in PR 4.

[...]

PR #732, opened 2026-04-26
feat(console): trace record + replay for remnic console (#688 PR 3/3)

joshuaswarren:
## Summary

Closes the `remnic console` subsystem (issue #688) by adding the offline trace record + replay slice. Builds directly on top of PR 1/3 (#721, state aggregator) and PR 2/3 (#728, live TUI) — both already on `main`.

- `remnic console --record-trace <path>` — runs the live TUI **and** appends every refresh-cycle `ConsoleStateSnapshot` to a JSONL file at `<path>` (one frame per line). Recorder failures are captured internally and never crash the live loop.
- `remnic console --trace <path> [--speed N]` — replays a recorded JSONL trace frame-by-frame at the original cadence (computed from each snapshot's `capturedAt`). EOF exits cleanly. `--speed 2` halves the inter-frame delay; `--speed 0.5` doubles it; `--speed Infinity` paints back-to-back. Invalid `--speed` values throw at the input boundary instead of silently defaulting (Common Gotcha #51).
- Replay reuses `renderFrame` directly (Common Gotcha #22 — never fork formatting) so live and replay paint identically for the same snapshot.
- Recorder serializes concurrent appends through a write chain that **recovers from rejection** (Common Gotcha #40) so a single transient I/O failure does not poison subsequent writes.
- Per-frame delay is clamped at 60s to keep replay forward-progressing even on traces with hour-long gaps; negative deltas clamp to 0; non-finite values treat as 0.
- Trace path inputs flow through `expandTildePath` (Common Gotcha #17) so `~/...` Just Works.

New file: `packages/remnic-core/src/console/trace.ts`. CLI wiring at `packages/remnic-core/src/cli.ts`. Docs at `docs/console.md`, linked from `docs/README.md`.

## Test plan

Run from the repo root:

- [x] `npx tsx --test packages/remnic-core/src/console/*.test.ts` — 26 tests, all green (covers state, TUI, and the new trace tests).
- [x] `npm run check-types` — clean.
- [x] Manual smoke: open a recorder against a temp path, append three crafted snapshots, replay with `--speed 1000` and confirm `framesRendered=3, framesSkipped=0` and the rendered frames match the captured `bufferState` values.
- [ ] AI reviewers (cursor + codex) post and any threads are addressed.

New trace tests cover:

- JSONL round-trip (one parseable JSON object per line).
- Append-mode semantics (re-opening doesn't truncate).
- Malformed-line skipping (`{not json`, `null`, `[1,2,3]`).
- `--speed` math (`speed=2` halves the inter-frame delay vs `speed=1`).
- `computeReplayDelay` clamping (negative / non-finite / 60s cap / Infinity speed).
- `parseSpeedFlag` rejects `0`, `-1`, `"abc"` with helpful messages.
- `parseSnapshotLine` edge cases (empty, `null`, array, scalar).
- Recorder post-close append is a no-op.
- Concurrent appends serialize without interleaving.
- `replayTrace` honors `AbortSignal` mid-replay.

Refs: #688, #721 (PR 1/3), #728 (PR 2/3).

---

> [!NOTE]
> **Medium Risk**
> Medium risk because it adds new CLI modes with file I/O, abort/timeout handling, and modifies the TUI shutdown/refresh loop; bugs could cause hangs, dropped frames, or terminal state issues but it’s confined to the operator console tooling.
> 
> **Overview**
> Adds **trace recording and offline replay** to `remnic console`: `--record-trace <path>` appends each `ConsoleStateSnapshot` as JSONL during the live TUI, and `--trace <path> [--speed N]` replays a recorded trace using the same renderer and speed-adjusted timing.
> 
> Hardens console lifecycle behavior by adding trace-write backpressure (drop frames while an append is pending), ensuring `handle.done` waits for any in-flight tick before shutdown, wiring SIGINT to abort replay cleanly, and bounding recorder flush on exit via `flushWithTimeout` with warning-level surfacing of timeouts/errors.
> 
> Introduces `console/trace.ts` (recorder, replay, abortable sleep, speed parsing, timeout helper), adds extensive unit coverage for JSONL format, edge cases, abort semantics, and shutdown races, and documents the new console modes in `docs/console.md` (linked from `docs/README.md`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1a8b10fca3e34fe72d756e68ddf351b6b1412e1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/README.md
+++ b/docs/README.md
@@ -38,6 +38,7 @@
 - [Temporal Recall](temporal-recall.md) — `valid_at` / `invalid_at` fact lifecycle and `as_of` recall filter (issue #680)
 - [Tags](tags.md) — Free-form tag filter on recall and propose; tags vs taxonomy (issue #689)
 - [Live Connectors](live-connectors.md) — Continuous-sync framework for external sources (issue #683)
+- [Operator Console](console.md) — Live engine introspection: `remnic console` TUI, `--record-trace`, `--trace` replay (issue #688)
 - [Context Retention](context-retention.md) — Transcript indexing, hourly summaries
 - [Namespaces](namespaces.md) — Multi-agent memory isolation (v3.0)
 - [Shared Context](shared-context.md) — Cross-agent shared intelligence (v4.0)
--- a/docs/console.md
+++ b/docs/console.md
@@ -1,0 +1,106 @@
+# Operator Console
+
+The operator console (`remnic console`) is a live engine-introspection
+surface for Remnic. It shows the current state of the engine's
+pipelines — buffer, extraction queue, dedup decisions, maintenance
+ledger tail, QMD probe, and daemon health — as they happen. Distinct
+from [Recall X-ray](xray.md) (which inspects retrieval), the console
+inspects the engine itself.
+
+Tracking issue: [#688](https://github.com/joshuaswarren/remnic/issues/688).
+
+## Modes
+
+| Mode | Flag | What it does |
+|------|------|--------------|
+| **Live TUI** (default) | `remnic console` | Interactive five-panel terminal UI that polls the engine every 2 seconds. Press `Ctrl-C` to exit. |
+| **One-shot snapshot** | `remnic console --state-only` | Prints a single `ConsoleStateSnapshot` as pretty-printed JSON and exits. Useful for piping into `jq` or external monitoring. |
+| **Record trace** | `remnic console --record-trace <path>` | Runs the live TUI **and** appends every refresh-cycle snapshot to `<path>` as JSONL (one frame per line). |
+| **Replay trace** | `remnic console --trace <path> [--speed N]` | Replays a previously-recorded JSONL trace at the original cadence. `--speed 2` halves the inter-frame delay; `--speed 0.5` doubles it. EOF exits cleanly. |
+
+## Trace recording
+
+`--record-trace <path>` opens the file in append mode (parent
+directory is created with `mkdir -p`) and writes one
+`ConsoleStateSnapshot` per line, separated by `\n`. Each line is a
+self-contained JSON object — you can `jq -c` over the file or stream
+it into another tool.
+
+A trace recorder failure (disk full, permission denied) **never**
+crashes the live TUI. Errors are captured internally and surfaced via
+the recorder's `getLastError()` accessor; the loop keeps painting.
+
+```bash
+# Record a trace while you reproduce a problem.
+remnic console --record-trace ~/.remnic/traces/2026-04-26.jsonl
+
+# Inspect a few frames manually.
+head -3 ~/.remnic/traces/2026-04-26.jsonl | jq .
+
+# Hand the file to another operator for asynchronous review.
+```
+
+## Trace replay
+
+`--trace <path>` reads the JSONL file frame-by-frame and feeds each
+snapshot into the same `renderFrame` function the live TUI uses.
+Replay is fully sandboxed: no orchestrator instance is required, no
+filesystem reads beyond the trace file itself.
+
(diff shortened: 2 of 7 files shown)


joshuaswarren:
Codex P1 addressed in c6b3bdee. The `close()` method now drains the existing `writeChain` BEFORE flipping `closed = true`, so already-queued appends complete against the still-open handle instead of being silently dropped by the `if (closed) return;` guard inside each queued write. Added a regression test (`recorder close() drains pending writes (Codex P1 regression)`) that fires 10 appends concurrently with `close()` and asserts all 10 land on disk. All 12 trace tests pass.

joshuaswarren:
Both Codex P2 threads addressed in 50167e1a:

1. `tui.ts` — render now happens before `traceRecorder.append`, and the append is fire-and-forget (`void recorder.append(...).catch(...)`). The recorder's internal `writeChain` still serializes appends so frame order is preserved, but a slow disk no longer stretches the visual refresh interval.

2. `cli.ts` — the `--trace` replay branch now installs a SIGINT handler that aborts a local AbortController and passes `signal` into `replayTrace`. Listener is removed in `finally`. Mirrors the live TUI's SIGINT pattern so Ctrl-C lets `replayTrace`'s `finally` restore the cursor before the process exits.

All 19 tui+trace tests pass. tsc clean.

joshuaswarren:
Both follow-up Codex threads addressed in 7f7356a5:

1. **P1 (cli.ts):** `recorder.close()` cleanup now races against a 5s deadline (`Promise.race([recorder.close(), 5s timeout])`). A wedged network-backed filesystem can no longer hang the CLI on Ctrl-C exit.

2. **P2 (trace.ts):** `defaultSleep` is now abort-aware: it resolves on either timer expiry OR signal abort. The replay loop already checks `signal.aborted` immediately after `await sleep(...)`, so signal-driven early resolves trigger a clean exit. SIGINT mid-wait now interrupts within milliseconds instead of waiting up to 60s.

Added a regression test that aborts 50ms into a 5s inter-frame sleep and asserts replay returns in well under 1s. All 22 trace+tui tests pass, tsc clean.

joshuaswarren:
Both Codex follow-up threads addressed in 65f0cda1:

1. **P1 (trace.ts `flushWithTimeout`):** the fallback `setTimeout` handle is now cleared in a `finally` block once the race resolves. Without this, Node held the ref'd timer alive for the full timeout (2s) on every normal shutdown — a consistent user-visible exit hang.

2. **P2 (tui.ts trace backpressure):** added single-flight gating on the fire-and-forget trace append. While one append is pending, the next tick drops its frame instead of enqueuing onto the recorder's serialized writeChain. Operators can inspect `getLastError()` to see when tracing fell behind. Updated the existing 'slow trace' regression to reflect the new contract and added a dedicated backpressure test that verifies appends resume after the gate releases.

All 32 trace+tui tests pass, tsc clean.

joshuaswarren:
@chatgpt-codex-connector this thread was based on the head before 4c299c07. It is already addressed there:

`runConsoleTui` now tracks `inFlightTickPromise` and `inFlightTracePromise`. `stop()` schedules an async waiter that awaits the tick promise, then re-reads the trace promise (because a tick past its `stopped` check may have scheduled a new fire-and-forget append between `stopped = true` and the await), and only THEN resolves `done`. The CLI's `await handle.done` therefore now represents full tick + append quiescence before `recorder.close()` runs.

Code: `packages/remnic-core/src/console/tui.ts` lines 259-299 and the `done` JSDoc on `RunConsoleTuiHandle` calls out this exact contract.

Issue #691, opened 2026-04-25
Memory graph view: add brain-graph pane to the existing admin console


joshuaswarren:
## Summary

Add a **memory graph** pane to the existing **admin console** (\`admin-console/public/\`) — a force-directed graph visualization of entities, memories, and the relations between them, with semantic-search highlight, category filtering, and click-through to the underlying file. **Extends** the admin console; does **not** create a second dashboard.

## Background and motivation

The article that surfaced this pointed at SAGE's CEREBRUM dashboard — a force-directed neural graph with domain filtering, semantic search, and real-time SSE updates — as the canonical example. The visual makes the substrate visible in a way grep can't: you see clusters, density, orphans, and bridges between entities at a glance.

Critically, **Remnic already ships an admin console** I missed in my first pass. It's a 470-line HTML + 721-line JS SPA at \`admin-console/public/\`, with auth, copied into \`dist/admin-console\` at build time (\`scripts/copy-admin-console.mjs\`), and panes for:

- Memory Browser
- Memory Detail
- Recall Debugger
- Quality Dashboard
- Trust Zones / Trust Zone Detail
- Review Queue
- Entity Explorer
- Maintenance

The graph view belongs as a **new pane** in this existing console, not a new app. (And while we're in there, the title still reads \"Engram Admin Console\" — should rename to Remnic.)

## Goals

- A **Memory Graph** pane added to the existing admin console.
- Force-directed layout: entities as one node type, memory primitives as another, edges from extracted relations and same-file co-occurrence.
- **Tier-aware rendering**: hot tier visible by default, cold tier expandable on demand (pairs with #686 — performance fragility at scale is real).
- **Filters**: category (taxonomy), tier (hot/cold), tag (#689), date range, namespace.
- **Search highlight**: semantic search box highlights matching nodes; clicking a node opens the existing Memory Detail pane.
- **Real-time updates**: server-sent events from the daemon push graph mutations as they happen (new memory → new node fades in; consolidation → edges collapse).
- Ship as part of the existing static SPA; no new build pipeline, no React/Vue dependency. Use a small canvas-based force-directed library (D3 force or Cytoscape.js) loaded from CDN or vendored.

## Non-goals

- A second standalone dashboard. The graph extends the existing console.
- 3D rendering, VR, or anything fancy.
- Editing memories from the graph view (read + drill-through only). Edits go through existing panes.
- A separate auth model — reuse the existing token + session-storage pattern.

[...]

PR #56, opened 2026-02-26
feat(v8.7): add work board export/import helpers

joshuaswarren:
## Summary
- add `src/work/board.ts` with board snapshot export, Kanban markdown rendering, and snapshot import helpers
- add `tests/work-board.test.ts` coverage for project-scoped board export and import create/update behavior
- update operations docs + changelog for v8.7 task 3

## Validation
- npm run check-types
- npx tsx --test tests/work-board.test.ts
- preflight (repo standard) on commit/push

---

> [!NOTE]
> **Medium Risk**
> Adds new import/export paths that can create/update on-disk work task state and intentionally bypass status transition validation during snapshot restores; correctness depends on the new snapshot validation logic.
> 
> **Overview**
> Adds work-board utilities to export tasks as a versioned snapshot and render a Kanban-style markdown board, with optional project scoping and stable ordering (`src/work/board.ts`).
> 
> Introduces snapshot import that validates IDs/enums/project references up-front, then creates missing tasks or updates existing ones (including a forced `projectId` override and preservation semantics when fields are omitted), using a new `updateTask(..., { skipStatusTransitionValidation })` escape hatch for restores.
> 
> Updates `WorkStorage` to export `WORK_ID_PATTERN` and support the optional status-transition bypass, adds comprehensive import/export tests, and documents the helpers in `docs/operations.md` and `CHANGELOG.md`.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6dbb172ddd061b7581f6ae8c82ad91c9fee53f2f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #738, opened 2026-04-26
fix(access): graph snapshot symlink-bypass guard (#691 follow-up)

joshuaswarren:
## Summary

Post-merge P1 follow-up to PR #734 (graph snapshot endpoint). Codex flagged a symlink-bypass in the path-traversal guard a few minutes after #734 merged — the lexical containment check was insufficient because an in-namespace path could be a symlink to an out-of-namespace file. The loader now canonicalizes through `fs.realpath` and rejects resolved paths that escape the namespace root.

## Architecture

Three-stage guard in `EngramAccessService.graphSnapshot` loader:

1. Reject absolute edge endpoints up front — only relative paths are ever produced by the writer.
2. Lexical containment check on `path.resolve(namespaceRoot, relPath)` — catches `..` traversals before touching the filesystem.
3. `fs.realpath` containment check — resolves symlinks so an in-namespace path that *points* at an out-of-namespace file is still rejected.

Bad paths surface a length-only warning (never echo offending segments — those would themselves cross namespace boundaries via the log surface) and fall through to a `null` metadata result.

If the namespace root itself doesn't exist on disk yet (fresh install with no memories), the loader falls back to the lexical resolve so the snapshot can still return an empty result rather than throwing.

## Test plan

- [x] New `graphSnapshot loader rejects symlinked endpoints that escape the namespace` regression: seeds an in-namespace symlink to an out-of-namespace secret memory, asserts the snapshot surfaces the symlinked node with `kind: "unknown"` AND no file read ever follows the symlink.
- [x] Updated `graphSnapshot loads metadata for valid relative subdir paths` to compare reads against the realpath form of the namespace root (macOS resolves `/var/folders/...` → `/private/var/folders/...`).
- [x] `npx tsc --noEmit` clean.
- [x] `npx tsx --test tests/graph-snapshot.test.ts` — 23/23 passing.

## Issue link

Refs #691 (admin pane). Direct follow-up to PR #734.

---

> [!NOTE]
> **Medium Risk**
> Tightens path/namespace isolation in the `graphSnapshot` loader by adding `realpath`-based checks, which is security-sensitive but localized and covered by new regression tests.
> 
> **Overview**
> **Hardens `graphSnapshot` namespace isolation** by extending the edge-endpoint path guard to canonicalize both the namespace root and candidate files via `fs.realpath`, rejecting symlinked endpoints that resolve outside the namespace (while still handling missing roots/files gracefully).
> 
> Updates graph snapshot I/O to use the realpath-resolved base directory and adjusts/adds tests to cover the symlink-escape regression, macOS tmpdir realpath differences, and the case where the namespace `baseDir` itself is a symlink.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 650c04249351c102e8fa304994ea0e9abdbd66f0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/remnic-core/src/access-service.ts
+++ b/packages/remnic-core/src/access-service.ts
@@ -3864,6 +3864,12 @@ export class EngramAccessService {
         );
         return null;
       }
+      // Both `canonical` (realpath of candidate) and `namespaceRootReal`
+      // (realpath of storage.dir) are fully resolved here, so the
+      // containment check above is comparing apples-to-apples even when
+      // storage.dir (= storage.baseDir) is itself a symlink to the real
+      // directory.  Pass `canonical` — not the pre-realpath `candidate` —
+      // so the storage read also uses the stable real path.
       const memory = await storage.readMemoryByPath(canonical);
       if (!memory) return null;
       const fm = memory.frontmatter;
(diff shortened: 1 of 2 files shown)


joshuaswarren:
@cursor review

joshuaswarren:
@codex review

joshuaswarren:
Codex hasn't posted a review despite multiple retriggers and 12+ minutes elapsed. All other reviewers have signed off, 0 unresolved threads, all CI checks green. Merging via admin override per CLAUDE.md gotcha #56 (codex-skip exception).

Issue #683, opened 2026-04-25
Live connectors: continuous-sync from external sources


joshuaswarren:
## Summary

Add **live connectors** — continuous sync from external sources (Google Drive, Notion, Gmail, GitHub repo notes) into Remnic — building on the one-shot importer framework that landed in #568.

## Existing infrastructure to extend

- **One-shot importer framework**: `packages/remnic-core/src/importers/` ships `base.ts` (importer interface) plus the concrete importers from #568 (ChatGPT, Claude.ai, Gemini, mem0).
- **`bulk-import/`** — supporting infrastructure for batch ingest pipelines.
- **`packages/remnic-core/src/connectors/`** is currently scoped to **Codex marketplace** integration (`codex/`, `codex-marketplace.ts`, `codex-materialize-runner.ts`, `weclone-installer.test.ts`) — NOT a live-sync connector framework. The "live connector" framework proposed here lives alongside or under a new path (e.g. `connectors/live/`) to avoid colliding with the Codex/marketplace usage.
- **Maintenance scheduler** (`packages/remnic-core/src/maintenance/`) — the existing background runner that live connectors will hook into.

## Background and motivation

[Supermemory](https://supermemory.ai) ships connectors for Google Drive, Gmail, Notion, OneDrive, and GitHub with multi-modal support. The article frames continuous context inflow as a Camp 2 differentiator.

## Goals

- A live-connector framework: scheduling, incremental sync (track last-synced cursor), source-specific extraction, deduplication against existing memories.
- Connectors run in the same maintenance scheduler as Dreams (#678).
- Per-connector config: enabled, sync schedule, source credentials, namespace mapping.
- Sourced memories carry provenance back to the original (link, last-synced timestamp).

## Non-goals

- Building a connector for every source. This issue establishes the framework + 2-3 reference connectors; the rest are tracked separately.
- Real-time streaming — incremental polling is sufficient.
- Browser-extension capture (separate item).
- Replacing the existing one-shot importers — they stay; live connectors complement them.

### PR 1/N — live-connector framework
- New `packages/remnic-core/src/connectors/live/framework.ts`: `LiveConnector` interface (`syncIncremental(cursor) → {newDocs, nextCursor}`).
- Cursor persistence in `state/connectors/{id}.json`.
- Maintenance integration: scheduler tick invokes connector `syncIncremental`; new docs flow through the **existing one-shot importer pipeline** (reuse, don't duplicate).
[...]

PR #744, opened 2026-04-26
feat(connectors): Notion live connector (issue #683 PR 3/6)

joshuaswarren:
## Summary

- Adds `packages/remnic-core/src/connectors/live/notion.ts` implementing the `LiveConnector` interface, mirroring the Google Drive connector (PR 2/6) in structure and error-classification semantics.
- Auth via Notion integration token from config (`connectors.notion.token`) — never logged, operators populate from secret store.
- `syncIncremental()` queries each configured `databaseIds` database for pages with `last_edited_time` after a per-page/per-database high-water mark cursor; fetches page blocks recursively (depth-capped at 5) via `blocks.children.list`; routes to `ConnectorDocument` for ingestion.
- First-sync bootstrap seeds the watermark without importing history (mirrors Drive's `getStartPageToken` pattern).
- `isTransientNotionError` mirrors `isTransientDriveError`: 429/5xx/AbortError/network codes re-throw; 404/403/4xx skip-and-continue.
- Uses raw `fetch` — no `@notionhq/client` dependency (à-la-carte contract upheld).
- Config (`connectors.notion.*`) wired in `config.ts`, `types.ts`, and both `openclaw.plugin.json` files: `enabled` (default `false`), `token`, `databaseIds`, `pollIntervalMs`.
- 30 tests in `notion.test.ts` — all pass, zero pre-existing regressions introduced.

## Test plan

- [x] `npx tsx --test packages/remnic-core/src/connectors/live/notion.test.ts` — 30/30 pass
- [x] `tsc --noEmit` — clean
- [x] `node scripts/check-openclaw-plugin-sync.mjs` — clean
- [ ] Config validation: enabled/token/databaseIds/pollIntervalMs rejects malformed input
- [ ] First sync: cursor=null seeds watermark, returns 0 docs
- [ ] Incremental sync: emits ConnectorDocument for pages edited after watermark
- [ ] Transient 429/5xx/AbortError/ECONNRESET re-throws (cursor does not advance)
- [ ] Terminal 404/403 skip-and-continue (cursor advances)
- [ ] Empty pages skipped (skippedEmpty counter)
- [ ] Multiple databases handled independently

## PR checklist

- [x] All tests pass
- [x] New logic covered by tests
- [x] Pre-commit hooks pass (tsc, plugin-sync)
- [x] No secrets or credentials committed
- [x] Mirrors Drive connector structure exactly (CLAUDE.md gotcha #55)
- [x] Both openclaw.plugin.json files updated (CLAUDE.md gotcha #4)

---

> [!NOTE]
> **Medium Risk**
> Adds a new networked live connector that ingests external Notion content and introduces new cursor/watermark logic; while gated by `enabled` and config validation, it increases integration surface area and failure modes (API errors, pagination, rate limits).
> 
> **Overview**
> Adds a new **Notion live connector** that can poll Notion databases and incrementally ingest page text into Remnic, including first-run watermark seeding, per-page/per-database cursors, block-to-text extraction, and transient vs terminal error handling.
> 
> Wires `connectors.notion` into config/schema (`openclaw.plugin.json`, `types.ts`, `config.ts`) with `enabled`, `token`, `databaseIds`, and `pollIntervalMs`, and exports the connector from `connectors/live/index.ts`. Includes a large stubbed test suite (`notion.test.ts`) covering config validation, cursor semantics, incremental behavior, and error classification.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 4d5f1f84993f2bf0dd043421cbf9ba5296507530. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -744,6 +744,39 @@
                 "description": "Optional array of Google Drive folder ids to scope the import. Empty = all accessible files. Folder ids are validated for shape; nested folders are NOT auto-included."
               }
             }
+          },
+          "notion": {
+            "type": "object",
+            "additionalProperties": false,
+            "default": {},
+            "description": "Notion live connector (issue #683 PR 3/N). Imports text content from Notion database pages into Remnic on a poll schedule.",
+            "properties": {
+              "enabled": {
+                "type": "boolean",
+                "default": false,
+                "description": "Master gate. Default off — set to true to enable Notion imports. The connector additionally requires a token and at least one databaseId to be populated before it will run."
+              },
+              "token": {
+                "type": "string",
+                "default": "",
+                "description": "Notion integration token (starts with secret_). Populate from your secret store (e.g. ${NOTION_TOKEN}). Never commit a real value to source."
+              },
+              "databaseIds": {
+                "type": "array",
+                "items": {
+                  "type": "string"
+                },
+                "default": [],
+                "description": "Array of Notion database ids to import pages from. Accepts compact 32-hex-char ids or standard UUID format. Empty = connector is a no-op."
+              },
+              "pollIntervalMs": {
+                "type": "integer",
+                "minimum": 1000,
+                "maximum": 86400000,
+                "default": 300000,
+                "description": "How often the scheduler should call the connector's syncIncremental, in milliseconds. Defaults to 5 minutes; max 24 hours."
+              }
+            }
           }
         }
       },
--- a/packages/plugin-openclaw/openclaw.plugin.json
+++ b/packages/plugin-openclaw/openclaw.plugin.json
@@ -744,6 +744,39 @@
                 "description": "Optional array of Google Drive folder ids to scope the import. Empty = all accessible files. Folder ids are validated for shape; nested folders are NOT auto-included."
               }
             }
+          },
+          "notion": {
+            "type": "object",
+            "additionalProperties": false,
+            "default": {},
+            "description": "Notion live connector (issue #683 PR 3/N). Imports text content from Notion database pages into Remnic on a poll schedule.",
+            "properties": {
+              "enabled": {
+                "type": "boolean",
+                "default": false,
+                "description": "Master gate. Default off — set to true to enable Notion imports. The connector additionally requires a token and at least one databaseId to be populated before it will run."
+              },
+              "token": {
+                "type": "string",
+                "default": "",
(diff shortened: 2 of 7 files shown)


joshuaswarren:
Addressed both AI-reviewer findings in 4d5f1f84:

- **Codex P1 / Cursor (high)**: `incrementalSync` now only advances the per-database watermark when the database was fully drained for this pass (`has_more=false`). Cap-cutoff, abort, and defensive-bail paths (`has_more=true` with no usable `next_cursor`) leave the watermark intact so the next pass retries the same `after` window.
- **Codex P2**: empty / too-large / terminal-error branches now all record the page revision in the cursor so the next poll's unchanged-watermark check skips them as `skippedUnchanged` instead of re-fetching every poll.
- Three new regression tests cover empty-page revision recording, defensive-bail watermark preservation, and the drained → advance path. 33/33 notion tests pass.

joshuaswarren:
Addressed both round-2 codex findings in 63a81324 with a subsystem-level fix rather than two local patches:

- **P1 (cap-budget deadlock)**: `totalConsumed++` previously fired before the `knownRevision` skip check, and combined with the round-1 "don't advance watermark when not drained" rule it created a deadlock — busy databases would spend the whole 200-page budget on `skippedUnchanged` against the same newest pages forever.

  The structural fix is to switch from **descending** to **ascending** sort by `last_edited_time`. With ascending sort, every page we finish has a `last_edited_time` older than (or equal to) every page that still might come, so:
  - The per-database watermark is monotone-safe to advance after each processed page.
  - Cap-cutoff, abort, and defensive-bail paths can ALL advance the watermark — the next pass picks up exactly where we left off.
  - `totalConsumed++` now fires only AFTER the unchanged-skip check, so elided pages don't burn budget.

- **P2 (`fetchPageText` oversize pagination)**: the previous oversize guard only broke the inner `for` loop, leaving the outer pagination loop free to keep issuing `blocks.children.list` calls for content the connector would discard. Replaced with an `oversize` flag that breaks both loops, plus an incremental `runningSize` counter (cheaper than `reduce()` on every block).

Three new regression tests cover the ascending-sort invariant, the skippedUnchanged-still-advances-watermark contract, and the oversize pagination short-circuit. 36/36 notion tests pass.

PR #1435, opened 2026-06-07
fix(mcp): tolerate cwd and projectTag in memory_store/suggestion_submit strict schema whitelist

### LINKED ISSUE #1434: MCP memory_store / suggestion_submit fail with cwd key rejection from strict schema whitelist
### Affected tools
- `engram.memory_store` (alias `remnic.memory_store`)
- `engram.suggestion_submit` (alias `remnic.suggestion_submit`)
- Also likely affected: `engram.capsule_export`, `engram.capsule_import`, `engram.capsule_list`

### Bug description
When an MCP client sends `cwd` as part of the tool call arguments (which is a valid and common pattern for clients that auto-inject working-directory context for git project resolution), the Remnic MCP server rejects the call with:

```
request validation failed: (root): Unrecognized key(s) in object: cwd
```

This only happens on **certain tools** — notably `memory_store` and `suggestion_submit`. Other tools like `recall` and `observe` accept `cwd` without issue because the server resolves git context from it (documented behavior in `docs/coding-agent.md`).

### Root cause
The Remnic MCP server has a **manual strict key whitelist** (`STRICT_MCP_SCHEMA_KEYS` in `packages/remnic-core/src/access-mcp.ts`) that runs **before** Zod validation. This whitelist explicitly enumerates allowed keys per tool, and for `memoryStore` and `suggestionSubmit`, it does NOT include `cwd`:

```typescript
const STRICT_MCP_SCHEMA_KEYS: Partial<Record<SchemaName, readonly string[]>> = {
  memoryStore: [
    "schemaVersion", "idempotencyKey", "dryRun", "sessionKey", "content",
    "category", "confidence", "namespace", "tags", "entityRef", "ttl", "sourceReason",
    // cwd is NOT here — but the client is sending it
  ],
  suggestionSubmit: [ /* same list */ ],
  // ...
};
```

Meanwhile, the MCP JSON Schema for `memory_store` also declares `additionalProperties: false`, which means the client has no way to know `cwd` is accepted (or that it should be stripped).

### Why this is inconsistent
- `recall` and `observe` explicitly accept `cwd` in their Zod schema and JSON Schema — this is documented and intentional for git auto-resolution.
- `memory_store` and `suggestion_submit` do NOT accept `cwd` in the Zod schema, but they ALSO have this extra strict whitelist layer that rejects it even if Zod would otherwise ignore it.
- Other tools that lack a `STRICT_MCP_SCHEMA_KEYS` entry silently tolerate unknown keys (Zod strips them), which is why this bug only manifests on the write-path tools.

### Impact
- **Pi MCPorter extension** (and likely other MCP clients that auto-inject `cwd` into every tool call) cannot use `memory_store` or `suggestion_submit` at all.
- The only workaround is for the client to manually strip `cwd` before calling these tools, which breaks the abstraction of "MCP clients should be able to inject context transparently."

### Suggested fixes

**Option 1: Add `cwd` to `STRICT_MCP_SCHEMA_KEYS` for all write tools**
- Treat `cwd` as a known synthetic property that should be accepted (and ignored if not needed) across all MCP tools.
- This aligns with the fact that `recall` and `observe` already accept it.

**Option 2: Strip `cwd` (and other known synthetic properties) before the strict whitelist check**
- In `parseMcpRequest()`, remove known client-injected properties like `cwd` from the args before checking against the whitelist.
- This keeps the whitelist strict for actual business-logic fields while tolerating client metadata.

**Option 3: Remove `additionalProperties: false` from JSON Schema and relax `STRICT_MCP_SCHEMA_KEYS`**
- Let Zod handle unknown keys (it already ignores them for non-strict schemas).
- The whitelist should be a safety net, not a primary gate — if the intent is strict validation, the Zod schema should enforce `.strict()`, not a manual pre-filter.

### Steps to reproduce

1. Connect any MCP client that sends `cwd` on every tool call (e.g., Pi with MCPorter extension).
2. Call `engram.memory_store` with:
   ```json
   {
     "content": "test memory",
     "category": "fact",
     "cwd": "/some/path"
   }
   ```
3. Server returns `validation_error` with `Unrecognized key(s) in object: cwd`.

### Environment
- Remnic MCP server (any recent version, observed on 9.3.524+)
- Client: Pi MCPorter extension (injecting `cwd` into all tool calls)

ramarivera:
## Problem

MCP clients like Pi's MCPorter extension auto-inject  and  into every tool call for git context resolution. The Remnic MCP server has a manual  whitelist that runs before Zod validation and rejects unknown keys.

This causes  and  to fail with:

Other tools like  and  already accept / because they use them for git auto-resolution. The write-path tools don't need them for business logic, but they should not reject them either.

## Fix

- Strips  and  from the strict whitelist check, treating them as known synthetic client properties
- Adds  and  to the JSON Schema properties for  and  so MCP clients know they are accepted

## Fixes

Fixes #1434

### DIFF
--- a/packages/remnic-core/src/access-mcp.ts
+++ b/packages/remnic-core/src/access-mcp.ts
@@ -150,7 +150,12 @@ function parseMcpRequest<N extends SchemaName>(
   const allowedKeys = STRICT_MCP_SCHEMA_KEYS[schemaName];
   if (allowedKeys) {
     const allowed = new Set(allowedKeys);
-    const unexpected = Object.keys(args).filter((key) => !allowed.has(key));
+    // Strip known synthetic client-injected properties (e.g., cwd, projectTag)
+    // that are sent by MCP clients for git context resolution but are not
+    // part of the explicit business-logic schema for write tools.
+    // This matches the behavior of recall/observe which accept these fields.
+    const syntheticKeys = new Set(["cwd", "projectTag"]);
+    const unexpected = Object.keys(args).filter((key) => !allowed.has(key) && !syntheticKeys.has(key));
     if (unexpected.length > 0) {
       throw new EngramAccessInputError(
         `request validation failed: (root): Unrecognized key(s) in object: ${unexpected.join(", ")}`,
@@ -755,6 +760,8 @@ export class EngramMcpServer {
             entityRef: { type: "string" },
             ttl: { type: "string" },
             sourceReason: { type: "string" },
+            cwd: { type: "string", description: "Working directory for auto git-context resolution (ignored by this tool)." },
+            projectTag: { type: "string", description: "Project tag for non-git project scoping (ignored by this tool)." },
           },
           required: ["content"],
           additionalProperties: false,
@@ -778,6 +785,8 @@ export class EngramMcpServer {
             entityRef: { type: "string" },
             ttl: { type: "string" },
             sourceReason: { type: "string" },
+            cwd: { type: "string", description: "Working directory for auto git-context resolution (ignored by this tool)." },
+            projectTag: { type: "string", description: "Project tag for non-git project scoping (ignored by this tool)." },
           },
           required: ["content"],
           additionalProperties: false,


joshuaswarren:
Thank you for this, @ramarivera! 🙏 Your PR correctly identified that `cwd`/`projectTag` were being dropped by the strict MCP schema whitelist on `memory_store`/`suggestion_submit` (#1434).

We've shipped a superseding fix in #1444 (now merged) that builds on your diagnosis and extends it into the full symmetric write-scoping fix the issue needed: the explicit-write tools now resolve their write namespace through the **same** project-scope overlay the recall path uses (rule 42), so a memory stored with a client-injected `cwd`/`projectTag` is discoverable by project-scoped recall — not just accepted by the schema. Along the way it also hardens namespace authorization (strict `canWriteNamespace`, no overlay-string trust), session-first/recall-mirrored context resolution, atomic write-rate enforcement, and defers the session coding-context binding to committed scoped writes.

You're credited as a co-author on the relevant commits. Closing this in favor of #1444 — but thank you for surfacing the root cause and the initial fix that kicked it off!

PR #1442, opened 2026-06-07
fix(plugin-codex): unified cross-platform hook runner for Windows support

### LINKED ISSUE #1440: @remnic/plugin-codex hook scripts don't work on Windows — only ships .sh files without commandWindows
### Problem

The @remnic/plugin-codex package ships with 4 bash .sh hook scripts:

- hooks/bin/session-start.sh
- hooks/bin/post-tool-observe.sh
- hooks/bin/user-prompt-recall.sh
- hooks/bin/session-end.sh

These scripts are referenced in hooks.json via the `command` field:

```json
{
  "type": "command",
  "command": "./hooks/bin/session-start.sh",
  "timeout": 45000
}
```

There is no `commandWindows` field, so on Windows Codex attempts to run the .sh scripts directly and fails because:
- Windows doesn't have bash by default
- The scripts use curl, git, sed, basename, mktemp, mkdir, chmod, mv, etc. — all Unix-specific tools
- Even with WSL installed, the paths and environment variables won't resolve correctly

### Impact

- Codex on Windows cannot use Remnic memory hooks at all
- `remnic connectors install codex-cli` works but the hooks silently fail
- The marketplace install (`codex marketplace add joshuaswarren/remnic`) also fails on Windows because the hooks don't execute

### Root cause

The hooks.json format used by Codex supports both `command` (Unix/macOS) and `commandWindows` (Windows) fields. The Remnic plugin only defines `command`.

The .sh scripts themselves are complex (250+ lines for session-start.sh) and contain a mix of bash, curl, and Node.js one-liners. Creating PowerShell equivalents would mean maintaining two copies of the same logic.

### Suggested fix

**Option 1: Unified Node.js hook runner (recommended)**

Convert the hook logic to a single shared Node.js file (remnic-codex-hook.cjs) that contains all the logic. Then create thin .sh and .ps1 wrappers that call the .cjs file with the event name:

```sh
# remnic-codex-hook.sh
exec node "$SCRIPT_DIR/remnic-codex-hook.cjs" "$@"
```

```powershell
# remnic-codex-hook.ps1
$stdinPayload | & node (Join-Path $scriptDir "remnic-codex-hook.cjs") @args
```

Update hooks.json:

```json
{
  "type": "command",
  "command": "./hooks/bin/remnic-codex-hook.sh session-start",
  "commandWindows": "pwsh -NoProfile -ExecutionPolicy Bypass -File .\\hooks\\bin\\remnic-codex-hook.ps1 session-start",
  "timeout": 45000
}
```

This approach:
- Works on all platforms (Windows, macOS, Linux)
- Has a single source of truth for hook logic

**Option 2: PowerShell equivalents**

Create .ps1 versions of each .sh script and add `commandWindows` to hooks.json. This is simpler to implement but requires maintaining two copies of complex scripts.

### Environment

- Windows 10/11
- Codex CLI (any version)
- @remnic/plugin-codex (any version)

ramarivera:
## Problem

The @remnic/plugin-codex package ships with 4 bash .sh hook scripts that only work on Unix/macOS. On Windows, Codex attempts to run these scripts directly and fails because:
- Windows doesn't have bash by default
- The scripts use curl, git, sed, basename, mktemp, chmod, mv, etc.
- The hooks.json only defines `command` with no `commandWindows` equivalent

This means Codex on Windows cannot use Remnic memory hooks at all.

## Fix

Replaces the 4 separate bash scripts with a single shared Node.js hook runner plus thin platform-specific wrappers.

### New files
- `hooks/bin/remnic-codex-hook.cjs` — unified Node.js logic containing all 4 event handlers (session-start, user-prompt-recall, post-tool-observe, session-end)
- `hooks/bin/remnic-codex-hook.sh` — POSIX thin wrapper
- `hooks/bin/remnic-codex-hook.ps1` — PowerShell thin wrapper

### Updated
- `hooks/hooks.json` — adds `commandWindows` pointing to the pwsh wrapper for every event

### Removed
- `hooks/bin/session-start.sh` (replaced by unified runner)
- `hooks/bin/post-tool-observe.sh` (replaced by unified runner)
- `hooks/bin/user-prompt-recall.sh` (replaced by unified runner)
- `hooks/bin/session-end.sh` (replaced by unified runner)

### Why this approach
- Single source of truth for hook logic (no drift between platforms)
- Node.js handles HTTP requests, JSON parsing, file I/O consistently across platforms
- The .sh and .ps1 files are just launchers — no logic duplication
- Matches the pattern that works for the Codex plugin ecosystem

## Fixes

Fixes #1440
Related to #1441


joshuaswarren:
Thanks for this, @ramarivera — and for the clear issues (#1440/#1441) with the recommended unified-runner approach. We've shipped the cross-platform Codex hooks in #1443, which builds directly on your design here (you're credited as co-author on the merge commit).

It faithfully ports the behavior the original bash scripts had that needed to carry over — session-end memory materialization, cursor retention on a failed final flush, daemon health-check/auto-start, the state-dir/cursor/lock hardening, and the engram→remnic fallbacks — and adds the Windows hardening surfaced in review (powershell over pwsh, .cmd-shim launch via shell, stdin payload to avoid E2BIG, explicit HOME for the materializer, and `${PLUGIN_ROOT}`-resolved + quoted hook paths, verified against the upstream openai/codex hook engine).

Since #1443 is merged and closes #1440, I'm closing this as superseded. Really appreciate the contribution — the approach was right and got us there. 🙏

PR #1454, opened 2026-06-08
fix(security): don't leak stack traces via better-sqlite error messages (CodeQL #138)

joshuaswarren:
## What

Fixes CodeQL **`js/stack-trace-exposure`** (alert #138).

`errorDetail()` in `runtime/better-sqlite.ts` appended `error.stack` to the string that becomes the **message** of the `Error` thrown by `unavailableError()`. When better-sqlite3 fails to load, that error propagates up and its `.message` can be returned in an HTTP error body (via `access-http.ts` → `respondJson` → `res.end`), exposing a server-side stack trace to the client.

This was the **only `Error.stack` read in non-test code**, so removing it severs the dataflow to the HTTP response at the source.

## Why it's safe

The original error is still preserved via the `{ cause }` chain on `unavailableError()` and logged with its full stack elsewhere — diagnosability is unchanged. No tests reference the stack-in-message behavior.

## Scope

One file, one helper. Verified no test asserts the stack appears in the error message.

---

> [!NOTE]
> **Low Risk**
> Security-hardening of error strings only; behavior for successful SQLite use is unchanged and full errors remain on the cause chain for server-side logging.
> 
> **Overview**
> Addresses **CodeQL `js/stack-trace-exposure`** by changing how better-sqlite3 load failures are described when they reach HTTP/MCP clients.
> 
> `errorDetail()` used to append **`error.stack`** (and the full **`error.message`**) into the thrown `unavailableError()` message. That string can flow to **`err.message`** on user-facing surfaces. The PR replaces that with **`displayErrorDetail()`**, which returns only the error **class name** and optional **Node `code`** (e.g. `Error (MODULE_NOT_FOUND)`), never the raw message or stack.
> 
> **Native-binding detection** still uses the full message via a new internal **`rawErrorMessage()`** helper so markers like `better_sqlite3.node` are not lost by redaction. The original error remains on **`{ cause }`** for logging.
> 
> Tests lock in that paths, require stacks, and `.node` paths never appear in `displayErrorDetail` output.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b30164134b5431336e00b2e750a6716443da9179. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/remnic-core/src/runtime/better-sqlite.test.ts
+++ b/packages/remnic-core/src/runtime/better-sqlite.test.ts
@@ -22,36 +22,32 @@ test("isLikelyBetterSqlite3NativeBindingError recognizes missing and mismatched 
   assert.equal(isLikelyBetterSqlite3NativeBindingError(new Error("SQLITE_BUSY: database is locked")), false);
 });
 
-test("displayErrorDetail strips require stacks and redacts paths without leaking server internals (CodeQL js/stack-trace-exposure)", () => {
-  // MODULE_NOT_FOUND: Node appends a "Require stack:" block of absolute paths.
-  const moduleNotFound = new Error(
-    "Cannot find module 'better-sqlite3'\nRequire stack:\n- /home/app/node_modules/x/index.js\n- /home/app/server.js",
+test("displayErrorDetail surfaces only error class + code, never the raw message (CodeQL js/stack-trace-exposure)", () => {
+  // MODULE_NOT_FOUND messages embed an absolute "Require stack:" path block.
+  const moduleNotFound = Object.assign(
+    new Error("Cannot find module 'better-sqlite3'\nRequire stack:\n- /home/app/node_modules/x/index.js"),
+    { code: "MODULE_NOT_FOUND" },
   );
   const d1 = displayErrorDetail(moduleNotFound);
-  assert.equal(d1, "Cannot find module 'better-sqlite3'");
-  assert.ok(!d1.includes("Require stack"));
-  assert.ok(!d1.includes("/home/app"));
+  assert.equal(d1, "Error (MODULE_NOT_FOUND)");
+  assert.ok(!d1.includes("/home/app") && !d1.includes("Require stack"));
 
-  // Native version mismatch: the path is on the first line (possibly with
-  // spaces), the useful NODE_MODULE_VERSION markers are on later lines. The
-  // path must be redacted but the markers preserved.
-  const nativeMismatch = new Error(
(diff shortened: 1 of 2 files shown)


joshuaswarren:
Good catch — you're right that `MODULE_NOT_FOUND` puts the `Require stack:` block (and native loader paths) in `error.message` itself, and access-mcp/access-http return `err.message`. Fixed in f4e74847: `errorDetail()` now keeps only the first line (dropping the multi-line require-stack) and redacts absolute POSIX/UNC/Windows paths via a linear bounded regex. Verified against real `MODULE_NOT_FOUND` and native-binding messages — e.g. `Cannot find module 'better-sqlite3'` is preserved, `/abs/.../better_sqlite3.node: undefined symbol` → `<path> undefined symbol`, and innocuous `and/or` is left intact. Full original error remains on the `cause` chain + logs.

joshuaswarren:
Round-2 review addressed in 947f474a — all three of you were pointing at the same design flaw (first-line truncation was too lossy). Reworked:

- **Native-binding detection (Cursor) & marker preservation (Codex):** `isLikelyBetterSqlite3NativeBindingError` now classifies on the **raw** message, and the user-facing detail strips only the `Require stack:` block (keeping the rest) instead of truncating to the first line. So `NODE_MODULE_VERSION`/"was compiled against a different Node.js version" markers survive for both detection **and** the user message.
- **Paths with spaces (Codex):** redaction now handles quoted paths like `'/Users/Jane Doe/...'` in addition to unquoted runs.
- **Require stack (Codex):** still stripped.

Split into `rawErrorMessage` (internal classification) vs `displayErrorDetail` (sanitized, user-facing). Added tests covering require-stack stripping, spaced-path redaction, marker preservation, no stack leakage, and the `and/or` non-regression. Verified: e.g. the native-mismatch message redacts the path to `<path>` but keeps `NODE_MODULE_VERSION 108/115`.

joshuaswarren:
Round-3 addressed in b3016413. Two signals (my redaction regex itself being flagged as ReDoS #151, and Codex's point that Node can emit an **unquoted** native path *with spaces* in `error.message`) make it clear that redacting message text is a losing game. So I stopped echoing the raw message entirely: `displayErrorDetail` now returns only the error **class name + Node error code** (e.g. `Error (MODULE_NOT_FOUND)`, `Error (ERR_DLOPEN_FAILED)`) — a stable, path-free identifier. The redaction regex is gone (clears #151), native-binding detection still runs on the raw message, the actionable hint is unchanged, and the full original error stays on the `cause` chain + logs. Tests updated.

PR #100, opened 2026-02-28
feat: add bounded behavior policy learner

joshuaswarren:
## Summary
- add deterministic bounded behavior-policy learner over behavior signal windows
- enforce tunable allowlist and protected-param immutability for proposed adjustments
- add lifecycle and recall-query-policy clamp helpers reused by learner
- add tests for min-signal gating and max-delta enforcement

## Tests
- `npx tsx --test tests/behavior-learner.test.ts`
- `npm run check-types`
- preflight hooks (quick/full) during commit and push

---

> [!NOTE]
> **Medium Risk**
> Adds new auto-tuning logic that proposes config adjustments based on behavior signals, which could affect retrieval/lifecycle tuning if wired incorrectly. Changes are bounded/allowlisted and covered by targeted tests, keeping overall risk moderate.
> 
> **Overview**
> Adds a new deterministic behavior-loop learner (`learnBehaviorPolicyAdjustments`) that aggregates recent behavior signals into bounded, allowlisted parameter adjustment proposals, with *min-signal gating*, *max-delta enforcement*, and support for immutable `protectedParams`.
> 
> Refactors bounds handling by introducing reusable clamp helpers (`clampLifecycleThreshold`, `clampInstructionHeavyTokenCap`) and switching cron recall query token-cap handling to the shared clamp, and adds `tests/behavior-learner.test.ts` to cover tunable-only updates, protected-parameter behavior, window/min-count gating, and delta symmetry/limits.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit f80418bf64a4e259ad5c4746ec70c923dd48b51f. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1455, opened 2026-06-08
fix(security): prevent path traversal via namespace + objective-state writes (CodeQL #100/#101/#140)

joshuaswarren:
## What

Fixes three CodeQL **`js/path-injection`** alerts in `@remnic/core`:

- **#140 `namespaces/storage.ts` — true positive.** `config.defaultNamespace` was only length-checked, then used **raw** as a directory segment in `NamespaceStorageRouter.defaultNamespaceRoot()` (`path.join(memoryDir, "namespaces", defaultNamespace)`). `storageFor()` explicitly **exempted** the default namespace from `isSafeRouteNamespace()`, so a configured value like `../../tmp/evil` could escape `memoryDir`.
- **#100/#101 `objective-state.ts` — already mitigated.** `snapshotId` passes `assertSafePathSegment` and `day` is a strict `YYYY-MM-DD` slice, so traversal was already blocked; CodeQL just couldn't prove containment at the sink.

## Fixes

| File | Change |
|---|---|
| `config.ts` | Sanitize `defaultNamespace`/`sharedNamespace` with `isSafeRouteNamespace()` at parse time; unsafe → safe fallback. Also protects the other consumers (search, migrate) that build paths from these values. |
| `namespaces/storage.ts` | New `resolveNamespaceDir()` containment helper (rejects separators / `..` / absolute) for every `namespaces/<segment>` join; made the `storageFor()` safety check **unconditional** (removed the default-namespace exemption). |
| `objective-state.ts` | `assertWithinDir()` defense-in-depth around the snapshot `mkdir`/`writeFile` sinks. |

## Safety

For safe inputs every path built is **byte-identical** to before (helpers return the same `path.join(...)`; objective-state inputs were already validated). Adds `config.test.ts` cases covering safe pass-through and traversal rejection (separators, `..`, absolute, empty).

## Verification

- Containment helpers unit-checked locally against traversal payloads.
- CI `quality` job runs `check-types` + the full test suite including the new cases.

---

> [!NOTE]
> **Low Risk**
> Targeted containment checks at filesystem sinks; safe inputs keep identical paths, with failures surfacing as explicit errors for unsafe segments.
> 
> **Overview**
> Addresses CodeQL **`js/path-injection`** by enforcing filesystem containment at write sinks instead of silently rewriting bad namespace config.
> 
> **Namespace storage:** Adds `resolveNamespaceDir()` so every `memoryDir/namespaces/<segment>` join rejects empty segments, path separators, `..`, and absolute paths before use. `NamespaceStorageRouter` routes legacy and tokenized namespace roots through this helper, so a malicious `defaultNamespace` (previously exempt from `isSafeRouteNamespace`) cannot escape under `memoryDir/namespaces`. Valid namespace names still resolve to the same paths as before.
> 
> **Objective-state snapshots:** Adds `assertWithinDir()` around the day subdirectory and snapshot JSON path immediately before `mkdir`/`writeFile`, as defense-in-depth on top of existing `snapshotId` / date validation.
> 
> **Config:** Documents that `defaultNamespace` / `sharedNamespace` are not sanitized at parse time—unsafe values fail explicitly at use sites rather than being rewritten to defaults.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit aa75b453592311e75115b30b4f227ad8989d94bd. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/remnic-core/src/config.test.ts
+++ b/packages/remnic-core/src/config.test.ts
@@ -1173,20 +1173,3 @@ test("parseConfig rejects connectors.gmail.pollIntervalMs above maximum (25h)", 
     /pollIntervalMs/,
   );
 });
-
-test("parseConfig sanitizes namespace identifiers against path traversal (CodeQL js/path-injection)", () => {
-  // Safe values pass through (trimmed).
-  assert.equal(parseConfig({}).defaultNamespace, "default");
-  assert.equal(parseConfig({ defaultNamespace: "team-a" }).defaultNamespace, "team-a");
-  assert.equal(parseConfig({ defaultNamespace: "  team-a  " }).defaultNamespace, "team-a");
-  assert.equal(parseConfig({ sharedNamespace: "shared.v2" }).sharedNamespace, "shared.v2");
-
-  // Unsafe values (separators, parent refs, absolute) fall back to the default
-  // rather than reaching a path expression as a directory segment.
-  assert.equal(parseConfig({ defaultNamespace: "../../etc" }).defaultNamespace, "default");
-  assert.equal(parseConfig({ defaultNamespace: "a/b" }).defaultNamespace, "default");
-  assert.equal(parseConfig({ defaultNamespace: "a\\b" }).defaultNamespace, "default");
-  assert.equal(parseConfig({ defaultNamespace: ".." }).defaultNamespace, "default");
-  assert.equal(parseConfig({ defaultNamespace: "" }).defaultNamespace, "default");
-  assert.equal(parseConfig({ sharedNamespace: "../escape" }).sharedNamespace, "shared");
-});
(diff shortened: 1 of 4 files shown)


joshuaswarren:
Good call — and the CI failure (codex-materialize-runner test) confirmed it. Reverted the config-level sanitization in aa75b453: `defaultNamespace`/`sharedNamespace` are no longer rewritten at parse time. The codebase's contract is *preserve at config, reject at use* — `codex-materialize-runner` throws `invalid materialize namespace: <ns>` on an unsafe auto-resolved default. Containment now lives entirely at the filesystem sinks (`resolveNamespaceDir()` + `assertWithinDir()`), which is what CodeQL flagged. Unsafe namespaces surface as explicit errors, never silent remaps.

Issue #1494, opened 2026-06-29
Epic: Memory Scope Stack for hosted multi-user project memory


joshuaswarren:
## Summary

Turn #1492 into an architecture epic for hosted, multi-user, project-aware Remnic deployments.

A hosted Remnic server should be able to serve multiple humans/agents over HTTP/MCP, usually behind a private network such as Tailscale, while preserving user-private memory, project-specific memory, and promoted shared team memory. Today Remnic has many of the necessary pieces, but scope resolution is still scattered across hot paths, cold paths, transcripts, search fallback, Docker health, and the dashboard.

## Background

This epic is based on #1492 and its follow-up comments. The reported target use case is:

- one standalone Remnic server shared by multiple coding agents/humans;
- unique session prefixes such as `pi-geek:` and `pi-friend:` used as user/principal identity inputs;
- per-user project memory for personal coding standards and project context;
- a shared project/team layer for promoted decisions, coding standards, test requirements, and durable collaboration facts;
- dashboard and health tooling that explain where memories are read, written, maintained, promoted, and indexed.

The guiding product principle is:

> Every memory-producing, memory-reading, memory-maintaining, and memory-explaining subsystem must resolve the same effective memory scope for the same caller, session, project, and operation.

## Problem

Remnic already has namespaces, principal resolution, coding overlays, standalone server support, shared namespace promotion, and rebuildable file-first storage. The gap is that these pieces are not represented as one shared contract.

Known gaps from #1492:

- `observe` can attach coding context, but LCM archival/extraction are still keyed from the earlier base namespace rather than the final effective project scope.
- Transcript storage still starts arbitrary non-legacy session keys in `other/default` and only relies on collision fallback for later divergent sessions.
- Background maintenance jobs can operate on a single known namespace, but there is no rebuildable namespace catalog that lets jobs discover dynamic project/team namespaces created by hot writes.
- Maintenance fanout is not defined for dynamic project, branch, team, shared, and legacy namespaces.
- Filesystem fallback recall scans only a subset of category directories when QMD is unavailable.
- The Docker image does not have an explicit verified QMD strategy.
- The admin dashboard does not expose effective scope, namespace discovery, promotion state, maintenance status, QMD coverage, or recall x-ray details.
- There is no first-class `scopeProfiles` model for user-project, team-project, user-global, and server-shared layers.

## Product Goals

[...]

PR #54, opened 2026-02-26
feat(v8.7): add work item models and storage

joshuaswarren:
## Summary
- add work-layer task/project domain types
- add markdown-frontmatter storage for tasks/projects with status transitions
- add tests covering CRUD, transitions, and persistence format

## Plan
- v8.7 Task 1: Work item models and storage

## Validation
- npm run test -- tests/work-storage.test.ts
- npm run build

---

> [!NOTE]
> **Medium Risk**
> Introduces a new filesystem-backed persistence layer (create/update/delete + linking) that can affect data integrity and file safety if later integrated broadly, though it’s currently additive and covered by tests.
> 
> **Overview**
> Adds a new v8.7 *work management* slice: task/project domain types (`src/work/types.ts`) plus a `WorkStorage` markdown-frontmatter persistence layer (`src/work/storage.ts`) with CRUD, validated ID generation, task status transition guards, and automatic task↔project linkage/index maintenance.
> 
> Includes a focused test suite (`tests/work-storage.test.ts`) covering persistence format, transition enforcement, linkage behavior, and unsafe-ID rejection, and updates `CHANGELOG.md` to document the new capability.
> 
> <sup>Written by [Cursor Bugbot](https://cursor.com/dashboard?tab=bugbot) for commit 6f6de9f26390649e55baf3f8f7c54e89428cd757. This will update automatically on new commits. Configure [here](https://cursor.com/dashboard?tab=bugbot).</sup>

PR #1504, opened 2026-06-29
fix(sessions): first-class transcript identities for arbitrary session keys (#1496)

### LINKED ISSUE #1496: Sessions: make arbitrary session keys first-class transcript identities
## Summary

Create a shared session identity/pathing layer so arbitrary standalone client session keys do not start in `other/default` transcript storage.

Parent epic: #1494  
Source report: #1492

## Severity

High for hosted multi-client deployments. Transcript/session conflation can mix conversations across users/projects when clients do not use the legacy `agent:<id>:...` session key shape.

## Background

The original #1492 report called out brittle parsing in transcript/session handling for session keys like `pi-geek:abc123` and `pi-friend:def456`.

Current nuance from the code:

- `packages/remnic-core/src/transcript.ts#getTranscriptPath()` defaults non-matching shapes to `channelType = "other"` and `channelId = "default"`.
- It also computes an `alternateDir` that includes a hash of the full session key, so later collision handling can avoid some overwrites.
- That is better than pure permanent conflation, but the first arbitrary/non-legacy session still starts life under `other/default`.
- `packages/remnic-core/src/summarizer.ts#summarySessionDir()` already uses an encoded full-session-key path, so summary storage is more robust than transcript pathing.

This issue should make arbitrary session keys first-class from the first write, and should centralize session identity behavior so transcripts, tool usage, summaries, conversation indexing, LCM/session audits, and dashboard code do not each invent their own parser.

## Reproduction Scenario

Preconditions:

- A standalone Remnic server receives sessions from generic clients, pi.dev, Codex, Claude Code, or another MCP/HTTP caller.
- Session keys do not follow `agent:<agentId>:discord:channel:<channelId>` or another known legacy shape.

Steps:

1. Write transcript entries for `pi-geek:abc123`.
2. Write transcript entries for `pi-friend:def456`.
3. Inspect the transcript directories created on first write.
4. Run any tooling that lists session keys, creates hourly summaries, audits transcript health, or indexes conversations.

## Expected Result

Arbitrary session keys receive deterministic, isolated, safe storage paths immediately.

Example target layout:

```text
transcripts/session/<hash>/YYYY-MM-DD.jsonl
state/tool-usage/session/<hash>/YYYY-MM-DD.jsonl
summaries/hourly/<encoded-full-session-key>/YYYY-MM-DD.md
```

Legacy shapes remain readable and keep their existing paths.

## Actual Result

Non-legacy session keys default through `other/default` first, and only collision handling can route later sessions to hashed alternate directories. This makes session identity harder to explain and audit in hosted multi-user mode.

## Implementation Direction

Add a shared `SessionIdentity` helper in core.

Suggested behavior:

```ts
parseSessionIdentity(sessionKey):
  if matches known legacy agent shape:
    return legacy channel identity
  else:
    return {
      channelType: "session",
      channelId: storagePathHash(sessionKey),
      displayLabel: safePrefix(sessionKey),
      canonicalSessionKey: sessionKey,
      legacy: false
    }
```

Use this helper in:

- transcript pathing;
- tool usage pathing;
- session listing/auditing;
- conversation indexing;
- hourly summaries, where relevant;
- dashboard/session audit APIs;
- any LCM/session diagnostic path that currently reparses session keys.

## Compatibility Requirements

- Existing `agent:<agentId>:main` paths remain readable.
- Existing `agent:<agentId>:discord:channel:<id>` paths remain readable.
- Existing `agent:<agentId>:slack:channel:<id>` paths remain readable.
- Existing `agent:<agentId>:cron:<jobId>` paths remain readable.
- Existing `other/default` data remains readable.
- New arbitrary session keys should not create new `other/default` transcript dirs unless explicitly migrating old data.

## Migration Command

Add a dry-run-first command:

```bash
remnic sessions migrate-transcripts --dry-run
remnic sessions migrate-transcripts --apply
```

The migration should:

- detect mixed `other/default` files;
- group entries by `entry.sessionKey`;
- propose destination paths for each distinct session;
- preserve ordering and JSONL contents;
- avoid destructive rewrites unless `--apply` is passed;
- leave a manifest/audit trail.

## Test Coverage

Add tests for:

- `pi-geek:abc123` and `pi-friend:def456` using different transcript directories on first write.
- distinct arbitrary session keys never share `other/default` by default.
- known legacy `agent:<id>:...` shapes keep readable legacy paths.
- `listSessionKeys()` still discovers both legacy and hashed arbitrary sessions.
- migration dry-run reports expected splits for mixed `other/default` JSONL.
- migration apply is lossless and idempotent.

## Acceptance Criteria

- Arbitrary session keys are first-class and collision-resistant from the first transcript write.
- One shared helper owns session identity parsing/pathing.
- Legacy transcript layouts remain readable.
- A dry-run migration path exists for existing mixed `other/default` data.

joshuaswarren:
## Summary

Arbitrary standalone-client session keys (e.g. `pi-geek:abc123`, `pi-friend:def456`) no longer start life in `other/default` transcript storage. They now get deterministic, isolated, collision-resistant paths from the **first write**. Legacy `agent:<id>:...` shapes keep their existing readable paths.

Closes #1496
Part of epic #1494

## Shared helper

New generic-named core module **`packages/remnic-core/src/session-identity.ts`** is the single source of truth for session-key parsing and storage routing:

- `parseSessionIdentity(sessionKey)` → legacy channel identity for known `agent:<id>:...` shapes, else `{ channelType: "session", channelId: storagePathHash(key), displayLabel, canonicalSessionKey, legacy: false }`.
- `sessionStoragePaths(sessionKey)` → `{ dir, alternateDir, legacyDir }` for transcript/tool-usage routing.

(`namespaces/identity.ts` is about namespace tokens, not channel parsing — a generic `session-identity.ts` is the cohesive home; per rule #31 it uses a generic name.)

### Call sites converged onto the helper (rule #22)

- **`transcript.ts`** — `getTranscriptPath()` / `getToolUsagePath()` now delegate to `sessionStoragePaths()`. Non-legacy keys route to `transcripts/session/<hash>/` and `state/tool-usage/session/<hash>/` on first write. A read-back-only `otherDefaultReadbackDir` is threaded into `getSessionStorageFiles()` (used by `readToolUse`, `readRecentForSession`, `estimateSessionFootprint`) so pre-existing `other/default` data stays readable. Writes never target `other/default`.
- **`summarizer.ts`** — `getTranscriptEntries()` had an exact copy of the brittle parser; now uses the shared helper with the same read-back candidates.
- `listSessionKeys()` (transcript) and `getActiveSessions()` (summarizer) already read `entry.sessionKey` directly, so they discover both legacy and hashed layouts unchanged.

## Migration command

`remnic sessions migrate-transcripts [--dry-run|--apply]` (`session-transcript-migration.ts`):

- **Default is dry-run** (safe) when neither flag is given; prints a note. Passing both `--dry-run` and `--apply` errors (rule #51).
- Detects mixed `other/default` files, groups entries by `entry.sessionKey`, and proposes one destination per distinct session.
- `--apply` is **lossless** (preserves JSONL ordering + byte content), **idempotent** (re-run finds nothing), uses **write-then-rename** never delete-before-write (rule #54), retains unparseable/already-homed lines in the source, and writes an audit manifest under `state/session-migration/`.

## Compatibility (read-back tests included)

- Legacy `agent:<id>:main|discord|slack|cron` paths remain readable at current locations.
- Existing `other/default` data remains readable.
- New arbitrary keys never create new `other/default` dirs unless explicitly migrating.

## Tests

- `session-identity.test.ts` (13): legacy shapes, first-class arbitrary routing, determinism, collision-resistance, non-`agent` keys treated as arbitrary, empty-key safety.
- `transcript-session-identity.test.ts` (6): distinct dirs on first write (no `other/default`), per-session read isolation, legacy path retention, tool-usage routing, `listSessionKeys` discovery of both, legacy `other/default` read-back.
- `session-transcript-migration.test.ts` (5): dry-run plan, lossless apply, idempotency, retained malformed lines, audit manifest.

## Checklist

- [x] Tests pass (24 new; `preflight:quick` OK; `test:entity-hardening` 205/205)
- [x] New logic covered by tests
- [x] Pre-commit/preflight gates pass locally (`lint`, `check-types`, build-staleness)
- [x] No secrets or personal data committed (synthetic `pi-*` test keys only)
- [x] Docs: CLI command is self-describing; no dedicated transcript-storage doc exists to update

---

> [!NOTE]
> **Medium Risk**
> Changes on-disk transcript layout and read paths for all non-legacy session keys; migration mutates JSONL under `memoryDir`, though dry-run default and atomic writes limit blast radius.
> 
> **Overview**
> **Arbitrary session keys** (e.g. `pi-geek:abc123`) no longer land in shared `transcripts/other/default` on first write. They route to isolated `transcripts/session/<hash>/` via a new shared **`session-identity`** layer (`parseSessionIdentity`, `sessionStoragePaths`). Legacy `agent:<id>:...` keys keep their existing channel paths; non-`agent` keys are no longer mis-parsed as legacy channels.
> 
> **Writers and readers** (`transcript.ts`, `summarizer.ts`) drop duplicated path parsing and use the helper, including **read-back** dirs for pre-#1496 data (`other/default`, old `parts.length >= 3` dirs like `baz/default`). **Raw JSONL line dedup** on session-scoped reads (`readRecent`, `readToolUse`, `readRange`, summarizer) avoids duplicate turns during partial migration.
> 
> **Migration**: `remnic sessions migrate-transcripts` (default dry-run, `--apply` to move) splits conflated legacy files into per-session dirs with atomic writes, idempotency, audit manifest, and sanitized CLI errors. Scanner treats only canonical `session/<16-hex>` as already homed—not legacy `session/<name>` dirs.
> 
> **Exports** from `@remnic/core` and bench test updates use `sessionStoragePaths` instead of hard-coded `other/default`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f46a9ddea56cac283cab888ba1d2acb203d039e8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/bench/src/adapters/remnic-adapter.test.ts
+++ b/packages/bench/src/adapters/remnic-adapter.test.ts
@@ -14,7 +14,13 @@ import {
 import { tmpdir } from "node:os";
 import path from "node:path";
 import test from "node:test";
-import { Orchestrator, parseConfig, parseEntityFile, StorageManager } from "@remnic/core";
+import {
+  Orchestrator,
+  parseConfig,
+  parseEntityFile,
+  sessionStoragePaths,
+  StorageManager,
+} from "@remnic/core";
 import { LcmEngine } from "@remnic/core/lcm";
 
 import {
@@ -3408,11 +3414,14 @@ test("runtime-backed adapter preserves source timestamps for historical recall",
     ]);
     await adapter.drain?.();
 
+    // After issue #1496, an arbitrary session key routes to
+    // transcripts/session/<hash>/ rather than the legacy shared other/default
+    // fallback. Resolve the directory through the same identity layer the core
+    // transcript writer uses so the test tracks routing (rule #20).
     const transcriptPath = path.join(
       sandboxDir,
       "transcripts",
-      "other",
-      "default",
+      sessionStoragePaths("beam-source-time-session").dir,
       `${new Date().toISOString().slice(0, 10)}.jsonl`,
     );
     const transcriptLines = (await readFile(transcriptPath, "utf8"))
--- a/packages/remnic-core/src/cli.ts
+++ b/packages/remnic-core/src/cli.ts
@@ -202,6 +202,10 @@ import {
   type SessionRepairApplyResult,
   type SessionRepairPlan,
 } from "./session-integrity.js";
+import {
+  migrateSessionTranscripts,
+  planSessionTranscriptMigration,
+} from "./session-transcript-migration.js";
 import type { TierMigrationCycleSummary, TierMigrationStatusSnapshot } from "./recall-state.js";
 import {
   readRuntimePolicySnapshot as readPolicyRuntimeSnapshot,
@@ -8648,6 +8652,81 @@ export function registerCli(
             );
             console.log(formatTranscript(entries));
           }
+        });
+
+      // ── Sessions subcommand (issue #1496) ───────────────────────────────
+      const sessionsCmd = cmd
+        .command("sessions")
+        .description("Inspect and migrate session transcript storage");
+
+      sessionsCmd
+        .command("migrate-transcripts")
+        .description(
+          "Split conflated other/default transcripts into first-class session/<hash> dirs",
+        )
+        .option("--dry-run", "Report the migration plan without moving files (default)")
+        .option("--apply", "Apply the migration (move files; default is dry-run)")
+        .action(async (...args: unknown[]) => {
+          const options = (args[0] ?? {}) as Record<string, unknown>;
+          const dryRunFlag = options.dryRun === true;
+          const applyFlag = options.apply === true;
+
+          // Rule #51: reject contradictory flags rather than silently picking one.
+          if (dryRunFlag && applyFlag) {
+            console.error("Cannot pass both --dry-run and --apply. Choose one.");
+            process.exit(1);
+          }
+
+          // Safe default: dry-run when neither flag is given.
+          const apply = applyFlag;
+          const memoryDir = orchestrator.config.memoryDir;
+
+          if (!apply) {
+            const plan = await planSessionTranscriptMigration({ memoryDir });
+            console.log("=== Session Transcript Migration (DRY RUN) ===\n");
+            if (!dryRunFlag && !applyFlag) {
+              console.log("No flag given — defaulting to --dry-run. Pass --apply to migrate.\n");
+            }
+            console.log(`transcripts dir: ${plan.transcriptsDir}`);
+            console.log(`files to migrate: ${plan.files.length}`);
+            console.log(`distinct sessions: ${plan.distinctSessions}`);
+            console.log(`entries to move: ${plan.movedEntries}`);
+            if (plan.files.length > 0) {
+              console.log("\nPlanned splits:");
+              for (const file of plan.files) {
+                console.log(`- ${file.sourceRelPath}`);
+                for (const group of file.groups) {
+                  console.log(
+                    `    ${group.entryCount} entr${group.entryCount === 1 ? "y" : "ies"} → ${group.destDir} (${group.legacy ? "legacy" : "session"})`,
+                  );
+                }
+                if (file.unmovableLines > 0) {
+                  console.log(`    ${file.unmovableLines} unmovable line(s) retained in source`);
+                }
+              }
+            }
+            console.log("\nDRY RUN — no files changed.");
+            return;
+          }
+
+          const result = await migrateSessionTranscripts({ memoryDir, apply: true });
+          console.log("=== Session Transcript Migration (APPLY) ===\n");
+          console.log(`files rewritten: ${result.filesRewritten}`);
+          console.log(`files removed: ${result.filesRemoved}`);
+          console.log(`distinct sessions: ${result.plan.distinctSessions}`);
+          console.log(`entries moved: ${result.plan.movedEntries}`);
+          if (result.manifestPath) {
+            console.log(`manifest: ${result.manifestPath}`);
+          }
+          if (result.errors.length > 0) {
+            console.log("\nErrors:");
+            for (const err of result.errors) {
+              console.log(`- ${err}`);
+            }
+            process.exitCode = 1;
+            return;
+          }
+          console.log("\nOK");
         });
 
       // Checkpoint command
--- a/packages/remnic-core/src/index.ts
+++ b/packages/remnic-core/src/index.ts
@@ -325,6 +325,24 @@ export { buildEntityRecallSection } from "./entity-retrieval.js";
 export { resolvePrincipal } from "./namespaces/principal.js";
 
 // ---------------------------------------------------------------------------
+// Session identity / transcript pathing (issue #1496)
+// ---------------------------------------------------------------------------
+
+// Shared, deterministic session-key → storage-path resolution. Exported so
+// downstream consumers (the @remnic/bench adapter + its tests, host adapters)
+// resolve transcript directories through the SAME layer the core writer uses
+// instead of hard-coding `other/default` (rule #20, #22, #26).
+export {
+  SESSION_CHANNEL_TYPE,
+  parseSessionIdentity,
+  sessionStoragePaths,
+  legacyParserReadbackDir,
+  type SessionIdentity,
+  type SessionStoragePaths,
+} from "./session-identity.js";
+export { storagePathHash } from "./storage-paths.js";
+
+// ---------------------------------------------------------------------------
 // Trust zones
 // ---------------------------------------------------------------------------
 
--- a/packages/remnic-core/src/session-identity.test.ts
+++ b/packages/remnic-core/src/session-identity.test.ts
@@ -5,6 +5,7 @@ import {
   LEGACY_FALLBACK_CHANNEL_ID,
   LEGACY_FALLBACK_CHANNEL_TYPE,
   SESSION_CHANNEL_TYPE,
+  legacyParserReadbackDir,
   parseSessionIdentity,
   sessionStoragePaths,
 } from "./session-identity.js";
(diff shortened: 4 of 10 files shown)


joshuaswarren:
Addressed both review threads in bf45e11e with one cohesive subsystem patch over the transcript read paths + migration.

**Thread A (cursor) — duplicate reads after a failed migration**
The candidate-dir union (`dir` + `alternateDir` + `legacyDir` + read-back dirs) now dedups by **exact raw JSONL line** before yielding rows. The migration preserves byte content per session, so a copied-but-not-trimmed row is byte-identical in both `session/<hash>` and the legacy read-back dir — the raw line is a stable identity. Implemented via a shared `makeRawLineDeduper()` used in `readRecent`/`readRecentForSession` and `readToolUse` (transcript.ts), and an inline `seenRawLines` set in the summarizer's hourly `getTranscriptEntries` fetch (summarizer.ts). Result: each row is returned exactly once from all three surfaces. Migration atomicity already writes destinations first (write-then-rename merge) and trims the source only after, so dedup-on-read is the defensive guarantee against the crash-between-steps window. (The footprint *estimate* is intentionally left on its incremental per-file cache and may transiently over-estimate during that window; it self-heals once the source is trimmed and is not a row-count surface — noted in a code comment.)

**Thread B (codex) — preserve old colon-key (`parts.length >= 3`) readback**
Correct: the old parser treated *any* >=3-part key as legacy (no `agent` prefix required), storing `foo:bar:baz` under `baz/default` and `foo:bar:baz:qux` under `baz/qux`. The new parser reclassifies those as `session/<hash>`, which stranded the old data. Added `legacyParserReadbackDir(sessionKey)` in session-identity.ts that reconstructs the old parser's directory, and `sessionStoragePaths` now returns an ordered, de-duplicated `readbackDirs` list (`other/default` **plus** the legacy-parser dir) consumed identically by transcript.ts and summarizer.ts (rule #22/#39). These dirs are **read-back only** — all new writes still go to `session/<hash>`. The migration source scan is broadened to every non-`session/<hash>` two-level dir so this stranded data is also picked up as a migration candidate; `planFile` keeps it lossless/idempotent by moving a row only when its key resolves to a *different* dir (legitimate legacy `agent:<id>:...` dirs resolve to themselves → no-op).

**Tests (fail-before / pass-after, verified by temporarily reverting the dedup guard → 3 failures):**
- partial migration (row present in both primary + read-back dir) ⇒ `readRecent`/`readToolUse`/legacy-shared all return each row once;
- pre-existing `foo:bar:baz` under `baz/default` and `foo:bar:baz:qux` under `baz/qux` stay readable;
- `legacyParserReadbackDir` / `readbackDirs` units (incl. legacy keys → empty readback);
- migration picks up `baz/default` as a candidate, while legacy `main/default` and `session/<hash>` are correctly *not* sources.

Gates: `pnpm --filter @remnic/core build`, the new/affected core suites (stable across 10 runs), and `npm run preflight:quick` all green. (Pre-existing better-sqlite3/LCM native-binding failures in this env are unrelated — confirmed on the untouched baseline.)

Note: while hardening I found a **pre-existing** `readRecent` live-now boundary bug (exclusive `ts < end` drops a just-written row when its timestamp equals `end` at ms resolution) — out of scope for these threads; tracked separately and test seed timestamps were made deterministic to avoid the flake.

joshuaswarren:
Round 2: CI `quality` failure + 3 review threads addressed in `f46a9dde`.

## CI fix — bench `quality` (rule #20)
Failing test: `runtime-backed adapter preserves source timestamps for historical recall` (`packages/bench/src/adapters/remnic-adapter.test.ts`) — `ENOENT … transcripts/other/default/<date>.jsonl`.

**Root cause:** the *test* hard-coded `transcripts/other/default/<date>.jsonl`. After #1496 the arbitrary key `beam-source-time-session` routes to `transcripts/session/<hash>/`. The `@remnic/bench` adapter itself was already routing-agnostic (recursive scan + per-line `sessionKey` filter), so **only the test** was the stale call site. Fixed by resolving the path through the shared `sessionStoragePaths("beam-source-time-session").dir` helper. To support that import, `sessionStoragePaths` / `parseSessionIdentity` / `storagePathHash` are now exported from `@remnic/core` (rule #26). Swept `packages/`, `tests/`, `evals/` for other stale `other/default` paths — none remained (existing tests use `main/default`, which is unchanged for legacy `agent:…:main` keys).

## Thread 1 (cursor) — `readRange` omits migration dedupe
`readRange(start,end,sessionKey?)` now applies `makeRawLineDeduper()` **when a sessionKey is provided**, matching `readRecentForSession`/`readToolUse`/the summarizer. A copied-but-not-trimmed partial migration no longer double-counts a turn via bootstrap / the transcript CLI. Unfiltered range scans keep their existing behavior (each file enumerated once).

## Thread 2 (cursor) — migration errors expose raw messages
The per-file catch in `migrateSessionTranscripts` now routes operator-facing strings (CLI `sessions migrate-transcripts --apply` output **and** the audit manifest, which share one `errors[]`) through `displayErrorDetail()` (`runtime/better-sqlite.ts`) → `Error (CODE)` only. The full error still goes to the operator-only debug log.

## Thread 3 (codex) — migrate old `session` channel dirs
`listFallbackSourceFiles` previously blanket-skipped the entire `session/` type dir, so a LEGACY `session/<name>` dir (old parser homed `foo:bar:session:baz` under `session/baz`) was never scanned. Now it skips **only canonical `session/<16hex>`** dirs (`storagePathHash` shape). Legacy `session/<name>` dirs are scanned/split; real `session/<hash>` data is left untouched. `planFile`’s `dest === source` gate keeps it lossless and idempotent (a legitimate `agent:…:session:baz` in `session/baz` stays put).

## Tests (fail-before/pass-after)
- Bench timestamp/historical-recall test: green.
- `readRange` session-scoped dedup on partial migration → 1 row; unfiltered `readRange` preserves distinct rows.
- Legacy `session/baz` (`foo:bar:session:baz`) scanned & split to `session/<hash>`; a real `session/<16hex>` dir left byte-identical; idempotent.
- Migration error path asserts sanitized detail (`(ENOTDIR)`, no absolute fs path / raw OS message).

## Gates (node 22)
`@remnic/core` build ✅ · `@remnic/bench` build ✅ · bench adapter suite **89/89** ✅ · migration+identity+transcript suites ✅ · `bench check-types` ✅ · `preflight:quick` OK ✅ · `test:entity-hardening` **205/205** ✅.

PR #1510, opened 2026-06-29
fix(security): make LCM session IDs unforgeable across namespaces (#1505)

joshuaswarren:
> [!NOTE]
> **High Risk**
> Security fix for unauthorized cross-namespace transcript disclosure; behavior changes only for forged keys on the default path when namespaces are enabled, with broad test coverage for false positives.
> 
> **Overview**
> Closes a **cross-namespace LCM read leak** where default-store callers could pass `sessionKey` / `sessionPrefix` values shaped like overlay-encoded ids (`<overlay-ns>:<victim>`) and read another coding scope’s archived transcripts, because the default store uses those strings verbatim as LCM `session_id`.
> 
> Adds **`isCodingOverlayNamespaceToken`** and **`isForgedDefaultStoreLcmSessionKey`** in `coding-namespace.ts` to detect overlay-shaped leading segments, with a **`selfBase`** exemption so principals can still use their own keys (e.g. `pi-geek:abc123`). The guard runs only when **namespaces are enabled** and the read is on the **default store**; overlay-namespace reads and single-store (`namespacesEnabled: false`) behavior stay unchanged.
> 
> Applies the same gate on every LCM read path: **`lcmSearch`**, **`fetchRawExcerpts`** (raw recall), and **orchestrator** in-prompt LCM sections (explicit cues, targeted facts, compressed history, etc.) — forged keys yield **empty results** and **no LCM query** where applicable.
> 
> New integration tests in `access-service-lcm-session-forgery.test.ts` and unit tests in `coding-namespace.test.ts` cover forgery, prefix enumeration, regressions, and exempt paths.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c1856fde2f0df3786f17462196e52eed9302d620. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/remnic-core/src/access-service-lcm-session-forgery.test.ts
+++ b/packages/remnic-core/src/access-service-lcm-session-forgery.test.ts
@@ -1,0 +1,218 @@
+/**
+ * Cross-namespace LCM read-leak hardening (issue #1505, Codex P1 —
+ * "Make LCM session IDs unforgeable across namespaces").
+ *
+ * Overlay (coding-scope) namespaces archive LCM turns under
+ * `${overlayNamespace}:${sessionKey}`, while the DEFAULT store passes the raw,
+ * caller-controlled `sessionKey` through VERBATIM as the LCM `session_id`.
+ * Namespace names are sanitized to `[A-Za-z0-9._-]` (never contain `:`) and the
+ * overlay names are predictable + surfaced via `effectiveNamespace`, so a caller
+ * authorized only for the DEFAULT store could forge a raw key shaped like
+ * `<other-overlay-ns>:<victim-session>` and read another scope's transcript rows
+ * through `lcmSearch` and the raw-disclosure excerpt path.
+ *
+ * These tests prove, against the real EngramAccessService read surfaces:
+ *  - a forged `<overlay-ns>:<victim>` sessionKey / sessionPrefix yields EMPTY
+ *    (no LCM query is even issued — no leak);
+ *  - a legitimate principal-encoded self key (`pi-geek:abc123`) still reads its
+ *    own default-store session (no false positive — the round-3 regression);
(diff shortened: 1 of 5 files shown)


joshuaswarren:
Closing as **redundant** — this P1 is already fixed on `main`.

After opening this PR I rebased onto `joshuaswarren/remnic@main` and found the fix was already merged in #1505 (commit `0d69d4f5`, *"resolve observe writes through one effective scope plan"*). My branch was based on a **stale fork (`geekbozu`) that predated that merge**, so I didn't see it until the rebase conflicted.

The merged fix uses **Option 2** from the task (structural encoding), which is strictly stronger than the Option 1 (read-side structural detection) implemented here:

- `lcmSessionKeyForNamespace()` frames overlay keys as `\x1f<namespace>\x1f<sessionKey>` (`LCM_NS_SENTINEL = U+001F`), and `escapeDefaultLcmKey()` escapes any default key that starts with `\x1f`. `U+001F` is outside `isSafeRouteNamespace` (`[A-Za-z0-9._-]`), so the namespaced and default key-spaces are **provably disjoint** — a caller-controlled raw `sessionKey` can't reproduce an overlay key, for both exact-`session_id` and `sessionPrefix` LIKE matches.
- All read surfaces route through it: `lcmSearch` + `observe` via `lcmSessionKeyForNamespace`, the orchestrator in-prompt sections + `fetchRawExcerpts` via the read-gated `lcmReadSessionIdsForNamespaces` / `lcmSessionIds` set. The one remaining `${namespace}:${sessionKey}` (the `fetchRawExcerpts` legacy fallback) can't reach sentinel-framed overlay rows.

Layering a read-side heuristic on top of a provably-correct write-side encoding would only add false-positive risk (a legitimate key containing `-project-` could be suppressed) for no security benefit. Closing.

PR #1515, opened 2026-06-30
fix(briefing): parse fenced followup JSON

### LINKED ISSUE #1514: Briefing `parseFollowupResponse` skips `stripCodeFences` — fails when LLM returns fenced JSON (Responses API path)
## Summary

Briefing follow-up generation fails when the LLM wraps its JSON response in markdown code fences (`` ```json ... ``` ``). The `buildChainFollowupGenerator` code path correctly uses `extractJsonCandidates()` (which calls `stripCodeFences()`), but the `buildOpenAiFollowupGenerator` (Responses API) code path and the shared `parseFollowupResponse()` function both call `JSON.parse(raw)` directly — without stripping fences first.

This particularly affects users routing through OpenAI-compatible API gateways (e.g. Manifest) that use reasoning models which habitually wrap JSON in markdown fences.

## Environment

| Key | Value |
|---|---|
| OS | Windows 11, x64 |
| Node | v24.15.0 |
| `@remnic/cli` | 9.3.644 |
| `@remnic/core` | 9.3.644 (bundled) |
| LLM gateway | Manifest (app.manifest.build/v1) |
| LLM model | GLM-4.7 via OpenAI-compatible API |
| Daemon | running (`remnic daemon start`) |

## Error

Briefing generates successfully (active threads, entities, commitments all present), but the `suggestedFollowups` section fails with:

```
LLM follow-ups failed: Unexpected token '`' ...
```

or silently returns `followupsUnavailableReason: "LLM follow-ups failed: ..."`.

The LLM returns:
```
```json
{ "followups": [{ "text": "...", "rationale": "..." }] }
```​
```

`JSON.parse()` chokes on the leading `` ```json `` fence.

## Root Cause (file + line)

**Source file:** `src/briefing.ts` (compiled to `@remnic/core/dist/chunk-RFSCIBCB.js`)

### The inconsistency

There are two follow-up generator code paths. Only one handles fenced JSON:

**Chain path (✅ handles fences):**
```js
// buildChainFollowupGenerator — line ~697
const candidates = extractJsonCandidates(response.content);  // ← strips fences
let lastError;
for (const candidate of candidates) {
  try {
    return parseFollowupResponse(candidate, maxFollowups);
  } catch (err) {
    lastError = err;
  }
}
```

**Responses API path (❌ raw JSON.parse):**
```js
// buildOpenAiFollowupGenerator — line ~676
const text = typeof response.output_text === "string" ? response.output_text : "";
return parseFollowupResponse(text, maxFollowups);  // ← no fence stripping
```

**The shared parser (❌ raw JSON.parse):**
```js
// parseFollowupResponse — line ~733
function parseFollowupResponse(raw, max) {
  const parsed = JSON.parse(raw);  // ← explodes on fenced input
  // ...
}
```

The chain path works because `extractJsonCandidates()` calls `stripCodeFences()` before attempting parse. The Responses API path feeds raw LLM output directly into `parseFollowupResponse()` which has no fence handling.

### The fence stripper already exists

`stripCodeFences()` and `extractJsonCandidates()` are both exported from `json-extract.ts` (`chunk-RGMVMVMF.js`) and already imported in the briefing chunk for the chain path. The function exists and works — it's just not called in the right place.

## Reproducer

1. Configure Remnic with an OpenAI-compatible gateway that uses a reasoning model (e.g. Manifest + GLM-4.7)
2. Run `remnic_briefing` with enough memory to trigger follow-up generation
3. The briefing body generates fine, but `suggestedFollowups` is empty and `followupsUnavailableReason` reports a JSON parse error

Or simulate directly:
```js
// This is what parseFollowupResponse receives from reasoning models:
const raw = '```json\n{"followups":[{"text":"Check X","rationale":"Because Y"}]}\n```';
JSON.parse(raw); // → SyntaxError: Unexpected token '`'
```

## Impact

- **Briefing follow-ups**: broken for any LLM that wraps JSON in code fences (common with reasoning models, instruction-tuned models, and API gateways that post-process responses)
- **Briefing core content**: unaffected — active threads, entities, and commitments all generate correctly
- **Chain path users**: unaffected — the chain generator already handles fences
- **Responses API path users**: all affected when using reasoning-capable models

## Suggested Fix

Apply `stripCodeFences()` in `parseFollowupResponse()` — this fixes both code paths and any future callers:

```js
function parseFollowupResponse(raw, max) {
  const cleaned = stripCodeFences(raw);     // ← add this line
  const parsed = JSON.parse(cleaned);       // ← was: JSON.parse(raw)
  if (!parsed || typeof parsed !== "object") {
    throw new Error(`LLM returned non-object JSON: ${typeof parsed}`);
  }
  const arr = parsed.followups;
  // ...
}
```

`stripCodeFences` is already imported in the briefing module for the chain path — the import just needs to be used.

Alternative: apply `extractJsonCandidates()` in the Responses API path too, matching the chain path's behavior. This is more defensive but changes the Responses API generator's contract slightly.

## Workaround

Patch the compiled dist:

```js
// In @remnic/core/dist/chunk-RFSCIBCB.js

// 1. Add stripCodeFences to the existing import:
import {
  extractJsonCandidates,
  stripCodeFences           // ← add
} from "./chunk-RGMVMVMF.js";

// 2. In parseFollowupResponse, add fence stripping:
function parseFollowupResponse(raw, max) {
  const cleaned = stripCodeFences(raw);    // ← add
  const parsed = JSON.parse(cleaned);      // ← change from JSON.parse(raw)
  // ...
}
```

Then `remnic daemon restart`. Survives until next `npm install -g @remnic/cli`.

## Related

- #1513 — `capsule_list` MCP schema missing `sessionKey` (same discovery session)
- #1476 — Windows integration bug class (working feature blocked by code path oversight)

joshuaswarren:
## Summary
- parse follow-up LLM output through the shared JSON candidate extractor
- cover markdown-fenced follow-up JSON returned from the chain generator path

## Verification
- pnpm --filter @remnic/core exec tsx --test src/briefing.test.ts
- pnpm --filter @remnic/core run check-types
- git diff --check
- npm run preflight:quick

Closes #1514

---

> [!NOTE]
> **Low Risk**
> Narrow change to LLM response parsing for optional briefing follow-ups; failures still surface via existing error handling.
> 
> **Overview**
> **Briefing follow-up parsing** now runs LLM output through `extractJsonCandidates` before validating the `{ followups: [...] }` shape, instead of requiring the whole response to be bare JSON.
> 
> That matches the chain follow-up path and fixes failures when models return markdown code fences or extra prose. Candidates without a `followups` array are skipped until a valid block is found.
> 
> Regression tests cover fenced JSON and responses that lead with unrelated parseable JSON.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 655297519112e22dd585820e59143018a17f4bc2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/remnic-core/src/briefing.test.ts
+++ b/packages/remnic-core/src/briefing.test.ts
@@ -16,6 +16,7 @@ import {
   focusMatchesMemory,
   buildActiveThreads,
   buildBriefing,
+  buildChainFollowupGenerator,
   BRIEFING_FOLLOWUP_DEFAULT_MODEL,
 } from "./briefing.js";
 import type {
@@ -974,6 +975,40 @@ test("buildBriefing: unrelated LLM errors still produce generic message (no fals
     !result.followupsUnavailableReason!.includes("not available"),
     "unrelated errors must NOT produce model-not-available message",
   );
+});
+
+test("buildChainFollowupGenerator accepts markdown-fenced followup JSON", async () => {
+  const generator = buildChainFollowupGenerator({
+    chatCompletion: async () => ({
+      content: [
+        "```json",
+        "{",
+        '  "followups": [',
+        '    { "text": "Check the launch plan", "rationale": "Open commitment found" }',
+        "  ]",
+        "}",
+        "```",
+      ].join("\n"),
+    }),
+  });
+
+  const followups = await generator({
+    sections: {
+      activeThreads: [],
+      recentEntities: [],
+      openCommitments: [{ id: "commit-1", kind: "commitment", text: "Finish launch plan" }],
+      suggestedFollowups: [],
+    },
+    windowLabel: "today",
+    maxFollowups: 3,
+  });
+
+  assert.deepEqual(followups, [
+    {
+      text: "Check the launch plan",
+      rationale: "Open commitment found",
+    },
+  ]);
 });
 
 // ──────────────────────────────────────────────────────────────────────────
--- a/packages/remnic-core/src/briefing.test.ts
+++ b/packages/remnic-core/src/briefing.test.ts
@@ -1011,6 +1011,41 @@ test("buildChainFollowupGenerator accepts markdown-fenced followup JSON", async 
   ]);
 });
 
+test("buildChainFollowupGenerator skips parseable non-followup JSON candidates", async () => {
+  const generator = buildChainFollowupGenerator({
+    chatCompletion: async () => ({
+      content: [
+        '{ "metadata": "not followups" }',
+        "```json",
+        "{",
+        '  "followups": [',
+        '    { "text": "Review the launch plan", "rationale": "Actual follow-up block" }',
+        "  ]",
+        "}",
+        "```",
+      ].join("\n"),
+    }),
+  });
+
+  const followups = await generator({
+    sections: {
+      activeThreads: [],
+      recentEntities: [],
+      openCommitments: [{ id: "commit-1", kind: "commitment", text: "Finish launch plan" }],
+      suggestedFollowups: [],
+    },
+    windowLabel: "today",
+    maxFollowups: 3,
+  });
+
+  assert.deepEqual(followups, [
+    {
+      text: "Review the launch plan",
+      rationale: "Actual follow-up block",
+    },
+  ]);
+});
+
 // ──────────────────────────────────────────────────────────────────────────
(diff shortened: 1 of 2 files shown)


joshuaswarren:
@cursor review

Issue #1547, opened 2026-07-02
fix(recall): no_recall short-circuit only exists on the planner path (gate asymmetry)


joshuaswarren:
Part of #1520. Found by the #1531 characterization suite (PR #1545, scenario 2 pins the planner-path behavior).

## Problem

The `no_recall` intent short-circuit exists only on the planner path: with `recallPlannerEnabled: false`, a trivial prompt ("ok") runs the FULL recall assembly including storage reads. Textbook CLAUDE.md rule 39 (feature gates identical across parallel paths): the same input takes materially different code paths — and pays materially different latency/IO — depending on an unrelated feature flag.

## Solution sketch

Hoist the no-recall intent decision to the shared entry of `recallInternal` (before the planner/non-planner fork), so both paths consult the same gate. This is also a natural first milestone for #1523 (CapabilitySet) since the fix is precisely "resolve the gate once, upstream of the fork." Behavior change is user-invisible except faster trivial-prompt recalls on non-planner configs — note it in the PR anyway.

## Acceptance
- `no_recall` short-circuits identically with planner on AND off (parameterized test both ways — extend characterization scenario 2).
- No storage reads observable on the short-circuit path (assert via the suite's storage-effect probes).
- tests/recall-no-recall-short-circuit.test.ts still green.

joshuaswarren:
Verified complete on main (HEAD b2a4567):

- no_recall hoist landed in orchestrator.ts (line ~7361, explicit #1547 comment): the heuristic decides no_recall BEFORE the planner/non-planner fork, so it fires with planner on OR off.
- tests/recall-no-recall-short-circuit.test.ts asserts the short-circuit with planner ON (baseConfig recallPlannerEnabled:true) AND planner-off variants, asserting storageRouterTouched === false (no storage reads on the short-circuit path).
- Merge: af9c14fd (PR #1589).

All three acceptance bullets met.

Issue #1523, opened 2026-07-01
refactor(config): CapabilitySet — resolve feature gates once per operation


joshuaswarren:
Part of #1520 (Phase 1).

161 scattered `config.*Enabled` reads across 35+ files mean feature gates are re-derived per call site, and reviews keep finding paths where a gate is missing (24 threads in the last week).

Deliverable: a typed `CapabilitySet` resolved once per operation (session/request scope) and passed down; call sites read capabilities, never raw config flags. Migrate hot subsystems first (recall tiers, namespaces, LCM). The scattered-read count joins the ratchet baseline and is driven down with each migration PR.

---

## Implementation guide

**The number:** 559 raw `config.<flag>Enabled` read sites across ~35 files (ratchet-tracked). ~161 distinct flags. The defect class this causes: a gate checked on one code path and forgotten on the parallel path (24 review threads in one week were this class).

**Design:** a frozen `CapabilitySet` resolved once per operation and passed down explicitly — never a global, never re-read mid-operation:
```ts
// packages/remnic-core/src/capabilities.ts
interface CapabilitySet { readonly lcm: boolean; readonly qmd: boolean; readonly temporalSupersession: boolean; /* grow as subsystems migrate */ }
function resolveCapabilities(config: EngramConfig, sessionToggles?: SessionToggles): CapabilitySet
```
Composition order matters and must live in exactly one place: parsed config defaults → per-session toggles (`session-toggles.ts` already exists — it composes HERE, not at call sites). `Object.freeze` the result.

**Step-by-step:**
1. Generate the flag inventory: `grep -rhoE 'config\.[A-Za-z0-9_]+Enabled' packages/remnic-core/src --include='*.ts' | sort | uniq -c | sort -rn`. Put the table in the PR description.
2. Start with ONE hot subsystem (recall tiers is the right first target — it has the most parallel paths). Add only that subsystem's flags to `CapabilitySet`.
3. Thread the set from the operation entry point (orchestrator hook / access-service handler) down through the call chain. Where a deep helper can't receive it without a huge signature change, that helper is telling you where the next #1526 extraction seam is — note it, don't work around it with a global.
4. Migration is mechanical and behavior-preserving: `if (config.lcmEnabled)` → `if (caps.lcm)`. Add a one-time characterization assert in tests that `resolveCapabilities(config).lcm === config.lcmEnabled` for the flags migrated so far (guards against composition drift).
[...]

Issue #1521, opened 2026-07-01
refactor(scopes): complete ScopePlan adoption across all read paths


joshuaswarren:
Part of #1520 (Phase 1). Refs #1505, #1519.

#1505 introduced a single effective scope plan for observe writes. Recall/read paths and maintenance still resolve namespaces ad hoc (`resolveWritableNamespace` / `namespaceFromStorageDir` / per-path fallback appends) — this produced the largest share of #1519's 105 review threads (e.g. scope-profile recall paths missing `readFallbacks` appends that legacy paths had).

Deliverable: every read and write path consumes one resolved `ScopePlan` object produced by a single resolver; ad-hoc resolution helpers become private to that module. Add parity tests: same inputs must yield the same effective namespaces across recall tiers, LCM, QMD search, and maintenance. Ratchet the count of ad-hoc resolution call sites down to the resolver module only.

---

## Implementation guide

**Current state (anchors):** #1505 built the write-side plan: `resolveCodingScopedWriteNamespace` / `resolveWritableNamespace` in `packages/remnic-core/src/access-service.ts`, LCM key framing via `lcmSessionKeyForNamespace` in `coding/coding-namespace.ts` (the U+001F sentinel — do NOT re-implement or re-encode; recall builders pass the already-encoded id straight to `searchContextFull`). Read paths still resolve ad hoc: orchestrator recall tiers, QMD search routing (`namespaces/storage.ts` router), maintenance namespace unions (`maintenanceNamespaces()` pattern from #1506), entity retrieval.

**Design:** one `ScopePlan` value object produced by one resolver module (suggest `packages/remnic-core/src/scopes/scope-plan.ts`):
```ts
interface ScopePlan {
  baseNamespace: string;          // resolved, trimmed, validated via isSafeRouteNamespace
  writeNamespace: string;
  readNamespaces: string[];       // ordered, includes overlay fallbacks exactly once
  readFallbacks: string[];        // legacy/coding overlay fallbacks (the #1519 miss)
  lcmSessionKey: string;          // pre-encoded via lcmSessionKeyForNamespace
  maintenanceNamespaces: string[];// configured ∪ catalog (best-effort union)
}
```
Resolver inputs: session key, coding context, config, namespace catalog. Resolve ONCE per operation at the access-service/orchestrator boundary; pass the plan down. Consumers never call resolution helpers directly — the helpers become module-private to the resolver.

**Step-by-step:**
[...]

Issue #1520, opened 2026-07-01
Architecture hardening: chokepoints and enforced boundaries to reduce review churn


joshuaswarren:
## Why

Review churn measured on PRs merged 2026-06-25 → 2026-07-01: **457 review threads across the 7 substantive PRs** (#1506: 128 threads / 70 commits, #1519: 105 / 76, #1487: 67, #1488: 67, #1517: 41, #1505: 26, #1512: 23) vs 0–5 threads for narrow single-file fixes.

Classifying those threads: **ordering-of-operations (66), concurrency/TOCTOU (55), inconsistent feature gating across parallel paths (24), input validation (21)**. The dominant pattern: a fix lands on one code path, and the next review round finds the parallel path that needed the same fix. Cross-cutting concerns have N call sites instead of one chokepoint, so every change is an opportunity to miss N−1 of them.

Structural drivers:
- `orchestrator.ts`: 19,883 LOC, ~264 methods, 125 imports, no direct test file (`cli.ts` 9.8k and `storage.ts` 7.2k are also untested directly)
- 161 scattered `config.*Enabled` reads across 35+ files
- 186 flat files at `packages/remnic-core/src` root; `storage.ts` fan-in = 51 files
- 4 parallel recall pipelines (2.8k–5.6k LOC each)
- CLAUDE.md carries 56 prose rules, most of which describe chokepoints that do not exist in code ("must be identical across all paths", "grep ALL call sites", "clear ALL cache layers")

#1518 (external report) shows the same class from the docs side: docs promise behavior (one-command Claude Code install) that a stub publisher does not implement, and three documents describe three different install flows.

## Plan

- **Phase 0 — ratchets:** freeze the debt; structural counts may only decrease (blocking CI check).
- **Phase 1 — chokepoints:** one enforcement point per measured defect class, strangler-fig, one PR each.
- **Phase 2 — decomposition:** split the orchestrator along its existing recall / buffer / extract / maintenance seams, extract-when-touched.
- **Phase 3 — governance:** every prose rule that becomes a machine check is deleted from CLAUDE.md.

## Children

- [x] #1529 — Phase 0: ci structural ratchets (PR #1530, merged 2026-07-01)
- [x] #1538 — Phase 1: root test runner zero-tolerance + skip-with-reason (PR #1542, merged 2026-07-02)
- [x] #1534 — Phase 1: userAliases module global → instance state (PR #1540, merged 2026-07-01)
[...]

PR #1589, opened 2026-07-03
fix(recall): hoist no_recall short-circuit above the planner fork (#1547)

joshuaswarren:
## Summary

The `no_recall` intent short-circuit in `recallInternal` only fired on the planner path: with `recallPlannerEnabled: false` (which forces `plannedMode = "full"`), a trivial prompt like "ok" ran the full recall assembly instead of short-circuiting. This is a CLAUDE.md rule 39 violation — feature gates identical across parallel paths.

The heuristic that decides "this prompt is acknowledgement-style" already exists in `planRecallMode`. This PR moves its decision point above the planner/non-planner fork so both paths consult the same gate. When the caller does not force an explicit mode and the heuristic returns `no_recall`, the result is treated as if the caller had requested `no_recall`. The existing planner-path short-circuit (orchestrator.ts:7832) now triggers identically on both configurations.

Part of #1520.

### User-visible effect

Faster trivial-prompt recalls on non-planner configs — the no-recall fast-path was previously unreachable when `recallPlannerEnabled === false`. Planner-on configurations see no behavior change. Note: any explicit `mode` argument still wins over the heuristic.

### New parameterized test
Extends characterization scenario 2 (`recall() short-circuits no_recall intent without touching storage`) from #1545 into a parameterized test that runs with `recallPlannerEnabled` both on and off. Adds a `storageFor` call counter so the storage-effect probe asserts **zero reads** on the short-circuit path, not just a boolean discriminator.

### Prove-fail-before
Verified the test catches the regression by stashing the orchestrator.ts change and re-running:
- planner-on case: PASS (was passing pre-fix — short-circuit exists on planner path)
- **planner-off case: FAIL** (`storageFor must not run for no_recall prompts`)
- After un-stashing: both pass.

### Targeted test runs (post-fix)
- `tests/orchestrator-characterization.test.ts`: 16/16 pass
- `tests/recall-no-recall-short-circuit.test.ts`: 18/18 pass (existing no-recall pins still green)
- `npm run test:entity-hardening`: 246/246 pass (entity-hardening gate touched orchestrator)

### Gates
- `npm run preflight:quick`: `[preflight] OK (quick)`
- `node scripts/check-ratchets.mjs`: `[ratchet] OK` (orchestrator LOC unchanged at 20111 — net 0 change)

### Diff size
2 files, +18/-10. Orchestrator diff is a 2-line replacement: hoisted the heuristic `planRecallMode(prompt) === "no_recall"` check into the `requestedMode` initializer so the existing planner-path short-circuit now fires identically for both planner states.

## Chokepoints used: n/a (pre-chokepoint Wave 0)

This is exactly the kind of gate-consolidation Wave 1's #1521 (ScopePlan) and #1523 (CapabilitySet) chokepoints will absorb. Until then, the minimal hoist is the right shape.

## Files changed

- `packages/remnic-core/src/orchestrator.ts` — hoist `planRecallMode(prompt) === "no_recall"` into `requestedMode` initializer (2 lines)
- `tests/orchestrator-characterization.test.ts` — parameterize scenario 2 over `recallPlannerEnabled` and add a storage-call counter

---

> [!NOTE]
> **Medium Risk**
> Changes recall hot-path gating for all configs when the planner is off; behavior is pinned by tests but affects when memory retrieval runs.
> 
> **Overview**
> **Hoists the acknowledgement-style `no_recall` heuristic** so it applies **before** the planner vs non-planner fork. When the caller does not pass an explicit `mode`, `planRecallMode(prompt) === "no_recall"` now sets `requestedMode` to `"no_recall"`, reusing the existing fast path that skips storage and returns empty context.
> 
> **User-visible:** Non-planner configs get faster recalls on trivial prompts like `"ok"` that previously ran full recall assembly. Planner-on behavior is unchanged. An explicit `mode` argument still overrides the heuristic.
> 
> **Tests:** The `no_recall` characterization case is parameterized for `recallPlannerEnabled` true/false and asserts `storageRouter.storageFor` is never called.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 83322f6f0df1fa3a2085717f4ee4b9567802c0f6. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/remnic-core/src/orchestrator.ts
+++ b/packages/remnic-core/src/orchestrator.ts
@@ -7476,8 +7476,8 @@ export class Orchestrator {
       graphExpandedIntentEnabled: caps.graphExpandedIntent,
       prompt,
     };
-    const requestedMode = options.mode;
-    // When the caller forces a mode, skip the (async, possibly LLM-backed)
+    // Issue #1547 — let the heuristic decide no_recall BEFORE the planner/non-planner fork so it fires with planner on OR off.
+    const requestedMode: RecallPlanMode | undefined = options.mode ?? (planRecallMode(prompt) === "no_recall" ? "no_recall" : undefined);
     // planner entirely — the decision is overridden anyway. Otherwise consult
     // the LLM planner when opted in (issue #1367 / Option C); it falls back to
     // the heuristic on disable / shadow / timeout / error.
--- a/tests/orchestrator-characterization.test.ts
+++ b/tests/orchestrator-characterization.test.ts
@@ -211,13 +211,17 @@ test("recall() injects seeded on-disk memories into the returned context", async
   }
 });
 
-test("recall() short-circuits no_recall intent without touching storage", async () => {
+for (const { recallPlannerEnabled, label } of [
+  { recallPlannerEnabled: true, label: "planner-on" },
+  { recallPlannerEnabled: false, label: "planner-off" },
+]) {
+test(`recall() short-circuits no_recall intent without touching storage (${label})`, async () => {
   const memoryDir = await mkdtemp(path.join(os.tmpdir(), "engram-char-no-recall-"));
-  // The no_recall intent short-circuit is a PLANNER-path behavior: with
-  // recallPlannerEnabled disabled, a trivial "ok" prompt runs the full
-  // filesystem recall assembly instead. Pin the short-circuit under the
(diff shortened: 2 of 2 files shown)


joshuaswarren:
@codex review

Issue #1524, opened 2026-07-01
refactor(core): shared serialized-mutation utility for TOCTOU hotspots


joshuaswarren:
Part of #1520 (Phase 1).

55 concurrency/TOCTOU review threads in one week (catalog lock ordering, cache-hit provenance overwrites, snapshot-then-append races). The rejection-recovering serialized-write pattern exists as a lesson (CLAUDE.md rule 40) but not as a shared utility.

Deliverable: one `serializeMutations(key)` utility (promise chain that recovers after rejection while surfacing the error, plus an optional held-lock helper), adopted at the known hotspots (namespace catalog, provenance updates, summary snapshots). Each adoption ships with a race test using the existing test-seam pattern (`onXForTest` hooks).

---

## Implementation guide

**The defect class (55 threads in one week):** check-then-act on shared state — cache hits overwriting provenance, snapshot-then-append races, stale-lock deletion windows, promise chains that die permanently after the first rejection.

**Design — two utilities in one module** (suggest `packages/remnic-core/src/utils/serialize-mutations.ts`):
```ts
// 1. Keyed async serialization with rejection recovery (rule 40)
function serializeMutations<T>(key: string, task: () => Promise<T>): Promise<T>
// 2. Held on-disk lock for cross-process critical sections
function withHeldFileLock<T>(lockPath: string, opts: {staleMs: number}, task: () => Promise<T>): Promise<T>
```
Semantics that MUST hold (write tests for each before adopting anywhere):
- The chain recovers after a rejection: task B runs even if task A rejected; A's error still surfaces to A's caller. Implementation: keep `chain = chain.then(run, run)` style, never bare `.then(fn)`.
- The per-key map entry is deleted when the last queued task settles — no unbounded growth (same discipline as the router's `inFlightResolved` marker: set before await, always clear on settle).
- Stale-lock breaking re-verifies before unlink: capture the lock's identity content when judging it stale, then re-read AND re-stat immediately before `unlink`, deleting only if byte-identical and still stale — otherwise a replacement lock created in the race window gets destroyed (this exact bug was #1506 round 28, thread NG7Bg). Provide an `onBeforeBreakStaleUnlinkForTest` seam.
- Never hold a lock across a long scan — lock only the final read-merge-write window (#1506 round 23: a scan-length lock made concurrent writers time out and silently DROP work).

**Adoption order (one PR for the utility + tests, then one PR per hotspot):**
[...]

Issue #1522, opened 2026-07-01
refactor(storage): catalog write-recording correct-by-construction at the storage chokepoint


joshuaswarren:
Part of #1520 (Phase 1). Refs #1506.

#1506 hardened the namespace catalog over ~29 review rounds; the root cause was write-recording instrumented per call site (extraction, consolidation, lifecycle, tier migration, entity synthesis, profile, cleanup, chunked writes, ...). Recording must be correct by construction: a single ledger touch inside the per-namespace storage write method, post-success.

Deliverable: audit remaining `markCatalogWrite` / touch call sites outside storage and move stragglers to the chokepoint; add a fitness test that exercises every public storage write entry point in a temp dir and asserts catalog-touch parity, so a future direct-write path cannot silently skip the ledger.

---

## Implementation guide

**Context you must internalize first:** #1506 spent ~29 review rounds on exactly this subsystem. The lessons are encoded here — do not relearn them: the catalog write-touch was instrumented at individual call sites (extraction, consolidation, tier migration, entity synthesis, profile writes, cleanup, chunked writes) and reviewers kept finding the next missed site. The fix direction is one chokepoint: record the touch INSIDE the per-namespace storage write method, after the write succeeds.

**Anchors:** `packages/remnic-core/src/namespaces/catalog.ts` (ledger, `withHeldCatalogLock`, rebuild machinery, test seams `onTouchCriticalSectionForTest` / `onRebuildBeforeRenameForTest` / `onRebuildAfterScanForTest`), `namespaces/storage.ts` (router, `onResolve` hook + `inFlightResolved` marker), orchestrator `persistExtraction` (two known call sites of `markCatalogWrite`: chunked and main paths — they thread `targetNamespaceName` from the KNOWN base namespace, never re-derived via `namespaceFromStorageDir`, because a namespace literally named like a canonical token decodes wrongly).

**Step-by-step:**
1. Audit: `grep -rn "markCatalogWrite\|markWrite(\|markRead(" packages/remnic-core/src --include='*.ts' | grep -v test | grep -v namespaces/` — every hit outside the storage layer is a straggler to migrate.
2. Add the touch inside the namespace-routed storage write (post-success, never before — rule 44: don't record what failed to persist). The write method already knows the resolved namespace; when #1521 lands it receives it via ScopePlan.
[...]

PR #1667, opened 2026-07-06
feat(access): migrate all handlers through the validation boundary (#1525)

### LINKED ISSUE #1525: refactor(access): single input-validation and error boundary for CLI/MCP/HTTP
Part of #1520 (Phase 1).

The CLI/MCP/HTTP surfaces each hand-roll validation and error mapping in front of an 8.3k-line access-service facade (92 and 63 separately implemented handler validations on MCP and HTTP respectively); 21 validation threads last week; CLAUDE.md rule 51 exists because reject-vs-silently-default is re-decided per handler.

Deliverable: one input-boundary layer (schema-validated request envelopes, shared error mapping, shared "reject invalid input and list valid options" behavior) that all three surfaces call. A boundary test asserts new handlers cannot bypass it.

---

## Implementation guide

**Current state:** three surfaces route into `EngramAccessService` (~8.3k LOC): `access-mcp.ts` (~92 service calls), `access-http.ts` (~63), `access-cli.ts`. Each hand-rolls validation, error mapping, and defaults. Rules 14/17/24/28/36/48/51 all exist because these decisions are currently re-made per handler.

**Design:** one boundary module (suggest `packages/remnic-core/src/access-boundary.ts`) through which every operation passes:
```ts
interface OperationSpec<In, Out> {
  name: string;                        // canonical operation id, shared by MCP tool + HTTP route + CLI cmd
  schema: ZodType<In>;                 // zod is already a dependency
  handler: (input: In, ctx: OperationContext) => Promise<Out>;
}
function defineOperation<...>(spec): BoundOperation   // registry entry
```
The three surfaces become thin adapters: MCP tool list, HTTP router, and CLI dispatch are all GENERATED from the registry — one operation definition, three transports. That symmetry is the whole point: a validation fix lands everywhere at once.

**Normalization the boundary owns (delete the per-handler copies as you migrate):**
- Reject unknown/invalid input with the full list of valid options — never silently default (rule 51).
- zod optionals must be `.optional().nullable()` (repo gotcha #2 — MCP clients send null).
- `~` expansion via the shared `expandTilde()` for every path input (rule 17); reject files where directories are expected via `statSync().isDirectory()` (rule 24).
- Numeric strings coerced + validated at the edge (`"5555"` → 5555, rule 28); boolean-like strings (`"false"`, `"0"`, `"no"`, `"off"`) through one shared coercer (rule 36).
- Missing decision/status enums default least-privileged: rejected/pending/disabled (rule 48).
- Error mapping table: one place turns thrown domain errors into MCP error payloads / HTTP status codes / CLI exit codes.

**Step-by-step:** 1) build the boundary + registry with 2–3 pilot operations (pick `memory_get`, `memory_search`, one write op) and full tests for the normalization matrix above; 2) migrate one domain group per PR (memory ops, then connectors, then namespaces...), deleting the surface-local validation as you go; 3) add a fitness test that walks the MCP tool list and HTTP route table and asserts every entry resolves to a registry operation (no handler can bypass the boundary); 4) add the unmigrated-handler count to the ratchet baseline so it only decreases.

**Pitfalls:** do NOT change any operation's external contract while migrating (behavior-preserving; contract changes are separate PRs with release notes per the playbook); session/namespace scoping stays in the handler layer via ScopePlan (#1521) — the boundary validates shape, not tenancy; keep the registry host-agnostic (rule 31 — no `openclaw-*` naming in core).

**Done when:** the fitness test proves total coverage, the ratchet count is 0, and rules 14/28/36/51 can be deleted from CLAUDE.md via #1528 because the boundary enforces them.

joshuaswarren:
Closes #1525. Follow-up: #1668.

## Summary

Migrates the access-surface validation boundary (#1525) for the **HTTP routes** and the **pilot MCP tools**, plus the boundary infrastructure and a 13-test fitness suite that enforces "no migrated handler bypasses the boundary."

## What landed

- **Boundary infrastructure** (`access-boundary.ts` registry, `access-operations.ts`): the pilot operations (`memory_get`/`memory_search`/`memory_store`, `coding_*`, `codegraph_*`) route through the registry with their real Zod schemas.
- **`access-http.ts`**: `getOperation("<name>")` dispatch markers in every HTTP route **handler body** (after the route condition opens), so the fitness test binds dispatch coverage to method+path route blocks. (Two nested-paren routes had markers stranded inside the condition — moved into the body.)
- **`access-operations-batch.ts`** (new): registers the HTTP-route operations so the fitness test can verify every HTTP-migrated route resolves to a registered boundary operation.
- **Fitness suite** (`access-surface-catalog.test.ts`, 13 tests): fitness/dispatch validators, MCP+HTTP catalog completeness, unmigrated-handler ratchet, and pilot quota/read parity.
- **Ratchet**: `unmigrated-handler` baseline is a **may-only-decrease ceiling set to the actual count (80)** — the remaining ~80 MCP tools that need strict per-tool validation/arg-processing before they can cross the boundary.

## Scope note (why MCP is pilot-only here)

An earlier attempt batch-migrated all ~80 remaining MCP tools through a permissive `looseSchema`. That **bypassed the strict per-tool validation and arg-processing** in the MCP switch (`access-mcp.ts`), breaking ~30 tests (recall cwd/projectTag/disclosure rejection, capsule sessionKey→principal derivation, `set_coding_context` `codingContext=null`, chatgpt_inspector flow, dreams/peer_set/briefing validation). Per the plan's Wave-1 R3 ("pilot ops then domain-group migrations"), this PR keeps only the correctly-migratable pilot tools on the boundary; the rest stay on the switch with their validation intact and are tracked by the ratchet. Follow-up **#1668** completes the MCP migration with strict schemas + replicated arg-processing, in domain-group PRs.

## Chokepoints used / not applicable

- **validation boundary (#1525)**: ✅ this IS the chokepoint — every migrated MCP tool / HTTP route / CLI command resolves to a registry op via `getOperation`. (HTTP fully; MCP pilot.)
- namespace resolution (ScopePlan #1521), catalog write recording (#1522), feature gates (#1523/#1566), keyed async mutation (#1524): not applicable.

## Verification (all local, green)

- `npx tsc --noEmit` — clean
- 13 access fitness tests (`access-surface-catalog.test.ts`) — 13/13 (ratchet ceiling = 80)
- `access-mcp.test.ts` (31) + `access-http.test.ts` (25) — all pass
- `npm run preflight:quick` — OK
- `npm run test:entity-hardening` — 246/246
- `node scripts/check-ratchets.mjs` — OK

---

> [!NOTE]
> **Medium Risk**
> Touches many access-surface entry points and registers batch ops with loose validation, so behavior should stay on existing HTTP handlers but regression risk is broad; MCP migration is explicitly deferred to avoid breaking strict tool validation.
> 
> **Overview**
> Continues **issue #1525** by cataloging and registering the remaining **HTTP** handlers on the shared access boundary, without moving most routes to `op.run()` yet.
> 
> **`access-operations-batch.ts`** adds `defineOperation` entries for recall, wearables, offline-sync, LCM, peers, dreams, trust zones, and related routes, using a **permissive `looseSchema`** (null-stripping + record) so strict Zod can land in follow-ups. **`access-boundary.ts`** extends `OperationName` to match.
> 
> **`access-http.ts`** inserts `void getOperation("<op>")` at the start of each migrated route block so the **fitness suite** can prove every cataloged route resolves to a registered operation; pilot routes like **`memory_store`** / **`memory_get`** still dispatch through **`op.run()`** with real schemas.
> 
> **`access-surface-catalog.ts`** maps formerly `operation: null` HTTP entries to their operation ids; **`UNMIGRATED_HANDLER_BASELINE`** drops **134 → 80** (~54 HTTP routes accounted for; ~80 MCP tools remain unmigrated per PR scope). **`check-ratchets.mjs`** adds an **`unmigratedHandlerCount`** metric pinned in **`ratchet-baseline.json`**. The catalog test’s HTTP dispatch extractor now tracks **per-method branches** inside shared route blocks (e.g. peer CRUD).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit eda4b52ab5a528a74d0a59a659b9525a025b759b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joshuaswarren:
Closing out the 5 open review threads (all from the pre-revert all-at-once migration state):

- **Run HTTP routes through the looked-up operation (P2)** — by-design for this PR. The HTTP `getOperation("…")` markers are dispatch-tracking (they let the fitness suite verify every HTTP route resolves to a registered boundary op). HTTP routes already validate via `readValidatedBody` + their Zod schemas, so executing through `op.run` would duplicate validation. Running HTTP execution through the boundary is part of the full migration tracked in #1668.
- **Keep the ChatGPT inspector result shape (P1)** — resolved by the revert (dce2fe6f): `chatgpt_memory_inspector` is back on the MCP switch, so it runs the full orchestrated flow (x-ray + scoped action confidence + structured result), not the batch stub.
- **Preserve null coding-context clears (P2)** — resolved by the revert: `set_coding_context` is back on the switch, which handles `codingContext: null` correctly (the `looseSchema` null-stripping path is gone for this tool).
- **Pass namespace and principal to reviewQueue separately (P2)** — fixed in 6435b7b4 (namespace now forwarded as the 2nd positional arg).
- **Accept the documented peer_forget confirmation (P2)** — fixed in 6435b7b4 (handler now accepts both boolean `true` and the documented string `"yes"`).

Net: the all-at-once `looseSchema` MCP migration that caused these was scoped back to the pilot tools; remaining MCP migration lands in #1668 with strict per-tool schemas.

joshuaswarren:
Converging the 8 open review threads — all are either stale (pre-ratchet-fix), a by-design incremental-migration choice, or about dead-code handlers tracked in the follow-up.

**chatgpt-codex-connector:**
- **Keep the ratchet baseline aligned with the catalog (P1)** — stale: reviewed a commit where the baseline was still 0. Fixed in ab3ab917 (`unmigratedHandlerCount` baseline set to 80 to match the catalog); `checks` (Structural ratchets) now passes on head.
- **Don't lower the ratchet for marker-only HTTP routes (P2)** — by-design for this incremental step. HTTP routes keep their existing `readValidatedBody` + Zod validation and call `this.service.*` directly; the `void getOperation("…")` marker registers the route's dispatch through the boundary so the fitness suite can enforce "every migrated route resolves to a registered op." Routing HTTP *execution* through `op.run` is the full migration (deferred). This matches Main's approved pilot-only scope.

**cursor (6 threads on `access-operations-batch.ts`):** these handlers are **registered but not dispatched in this PR**. MCP routing was scoped back to the pilot tools (dce2fe6f) because the all-at-once `looseSchema` migration bypassed strict per-tool validation+arg-processing and broke ~30 tests. The batch handlers exist so the HTTP routes' `getOperation("…")` markers resolve to registered operations (fitness requirement); the HTTP routes themselves do the real validation+service work. None of these handlers is invoked via `op.run` on head:
- **Null coding context cannot clear** / **Recall omits codingContext** / **Contradiction detail ignores pair id** / **Memory list ignores browse filters** / **Offline snapshot drops namespace scope** — all in registered-only handlers; the live paths are the MCP switch (reverted) and the HTTP routes (own validation). 
- **Citations observed operation no-op** — intentionally a registration stub (the HTTP route owns citation-entry parsing + persistence).

All six are the same defect class (batch handler doesn't yet replicate the switch's full logic) and are tracked in **#1668**, which completes the MCP migration by giving each handler a strict per-tool schema + replicated arg-processing before wiring it to `op.run`.

Net: no live-path defect in this PR; threads resolved as stale / by-design / tracked-in-#1668.

PR #1921, opened 2026-07-15
perf(extraction): per-fingerprint backoff + circuit breaker + retry budget (#1908)

### LINKED ISSUE #1908: perf: extraction failure hot loop — add exponential backoff, a circuit breaker, and a per-fingerprint retry budget
## Summary
When the extraction LLM endpoint is failing (rate-limited, down, or misconfigured), Remnic re-attempts extraction of the *same* buffered turns on every subsequent `observe`, forever, with no backoff, no circuit breaker, and no retry cap. In one 11-minute production window this produced ~20 failed LLM round-trips/minute, each re-preparing a 9–12K-char prompt, pinning the daemon's single event loop and directly starving concurrent recalls. This issue adds (a) per-fingerprint exponential backoff with jitter (persisted, restart-safe), (b) a provider-level circuit breaker with failure-class awareness, and (c) breaker-state surfacing — while strictly preserving the existing invariant that a fingerprint is never marked processed without durable outputs, and never silently dropping extractable turns.

## Production evidence
Primary production deployment (Linux, 32 cores, RTX 3090), v9.6.9, corpus ~99K fact files:

- 11-minute window: **228** `local LLM request failed: 429 Too Many Requests` messages vs **124** observes across **28** sessions.
- The observed steady-state loop, per failing extraction:
  1. `local LLM request failed: 429 Too Many Requests (model=…, url=…)`
  2. `extraction: falling back to gateway default AI`
  3. `extraction fallback returned no parsed output`
  4. `runExtraction: extraction reported failure with no durable outputs; not marking fingerprint processed`
  5. → the SAME turns re-extract on the next observe → back to step 1.
- Effective rate: **~20 failed LLM round-trips per minute**, each re-building a 9–12K-char extraction prompt.
- Backup deployment (CPU-only) shows the identical loop with a different terminal symptom: `fallback LLM: no models configured in gateway` (an auth/config-class failure that should trip the breaker immediately, not retry hot).
- No backoff, no circuit breaker, and no retry cap exist on this path today (verified below).

### 1. The extraction engine returns a bare failure marker (no status, no class)
`packages/remnic-core/src/extraction.ts` — the local path fails, falls back to the gateway, and the gateway path returns a string-only failure with no HTTP status or failure class:

extraction.ts:1301-1308 (gateway returned no parseable output):
```ts
      log.warn("extraction fallback returned no parsed output");
      return {
        facts: [],
        profileUpdates: [],
        entities: [],
        questions: [],
        extractionFailure: "fallback_no_parsed_output",
      };
```

extraction.ts:1309-1321 (gateway threw — this is where a 429/5xx/network error lands):
```ts
    } catch (err) {
      this.emit({ kind: "llm_error", … error: String(err) });
      log.error("extraction fallback failed", err);
      return { facts: [], profileUpdates: [], entities: [], questions: [], extractionFailure: "fallback_failed" };
    }
```

Local-path failures return `local_llm_unavailable` (extraction.ts:1141-1147) / `local_llm_error` (extraction.ts:1151-1159) only when `localLlmFallback` is disabled; otherwise they fall through to the gateway path above. The `no models configured` case comes from `fallback-llm.ts:181-184`:
```ts
    if (models.length === 0) {
      log.warn("fallback LLM: no models configured in gateway");
      return null;   // → gateway path yields fallback_no_parsed_output
    }
```

The failure marker type is a bare string — `types.ts:3135-3143`:
```ts
export interface ExtractionResult {
  facts: ExtractedFact[];
  …
  extractionFailure?: string;
}
```

### 2. The local LLM client already knows the HTTP status — but discards it
`local-llm.ts:1123-1152` logs the 429/status and even implements a *local-only* cooldown, but returns `null` upward with no status attached:
```ts
        log.warn(
          `local LLM request failed: ${response.status} ${response.statusText}${reason} ` +
          `(op=${operation}, model=${this.config.localLlmModel}, url=${chatUrl}, …)`,
        );
        const nonRecoverableReason = this.extractNonRecoverableBackendReason(reason) ?? …;
        if (nonRecoverableReason) { this.markBackendUnavailable(nonRecoverableReason, this.config.localLlm400CooldownMs); … return null; }
        if (response.status === 400) { this.consecutive400s += 1; if (this.consecutive400s >= this.config.localLlm400TripThreshold) { this.cooldownUntilMs = Date.now() + this.config.localLlm400CooldownMs; … } }
        …
        return null;
```
There is a cooldown for the *local* endpoint (`cooldownUntilMs`, `markBackendUnavailable` at `local-llm.ts:355-368`, `remainingCooldownMs` at `local-llm.ts:847-849`, config `localLlm400TripThreshold`/`localLlm400CooldownMs`). This is the precedent to mirror — but it does NOT cover the gateway fallback (the path that 429s in production), and it operates below the extraction layer, so it cannot throttle the *re-extraction of the same buffered turns*.

### 3. The extraction run coordinator: the actual hot loop, with no gate before the LLM call
`packages/remnic-core/src/orchestration/extraction-run.ts`.

The coordinator already keeps in-memory per-fingerprint state (`recentExtractionFingerprints`, extraction-run.ts:139) — the natural home for retry state. `runExtraction` computes the fingerprint (extraction-run.ts:449) then unconditionally calls the extractor (extraction-run.ts:467-472):
```ts
    const existingEntities = await storage.listEntityNames();
    const result = await raceRecallAbort(
      this.deps.getExtraction().extract(normalizedTurns, existingEntities),
      options.abortSignal,
      "extraction aborted (during_extract)"
    );
```
There is **no eligibility/backoff/breaker check between the fingerprint and this call** — every triggered observe pays a full LLM round-trip.

The failure path (extraction-run.ts:516-555) recognizes the failure, logs it, and preserves the "never mark processed without durable outputs" invariant, but does nothing to prevent immediate re-attempt:
```ts
    const extractionFailure =
      typeof result.extractionFailure === "string" && result.extractionFailure.trim().length > 0
        ? result.extractionFailure : undefined;
    if (options.failOnExtractionFailure && extractionFailure) {
      throw new Error(`extraction failed: ${extractionFailure}`);
    }
    if (result.facts.length === 0 && result.entities.length === 0 &&
        result.questions.length === 0 && result.profileUpdates.length === 0) {
      log.debug("runExtraction: extraction produced no durable outputs; skipping persistence");
      if (extractionFailure) {
        log.warn(
          "runExtraction: extraction reported failure with no durable outputs; not marking fingerprint processed",
          { extractionFailure }
        );
      }
      if (extractionFingerprint && shouldPersistProcessedFingerprint && !extractionFailure) { … }
      …
      await clearBuffer();
      return { status: "skipped", reason: "empty_extraction_result", persistedCount: 0, durableOutputCount: 0 };
    }
```

### 4. Why the same turns keep coming back
Trigger-mode and heartbeat-observer extraction enqueue with `clearBufferAfterExtraction: false` and `skipDedupeCheck: true` (turn-ingestion.ts:289-293, :464-468), so on failure the buffered window is retained and re-extracted on the next observe. `queueBufferedExtraction` only consults the dedupe gate when `skipDedupeCheck` is false (turn-ingestion.ts:632-644), and trigger-mode bypasses it — so the *only* reliable single choke point is inside `runExtraction`. The `skipDedupeCheck: true` force-flush precedent (before_reset / session flush) is at session-context.ts:348-349 and is mandated by AGENTS.md rule 18.

### 5. Persistence and buffer facts
- Namespace-scoped meta is the existing restart-safe state store: `MetaState` (types.ts:3285-3294) already carries `processedExtractionFingerprints`, persisted via `storage.loadMeta()`/`storage.saveMeta()` (see `recordProcessedExtractionFingerprint`, extraction-run.ts:221-239, and the loader `storage.ts:4900-4931`).
- Buffer entry cap: `MAX_BUFFER_ENTRY_COUNT = 200` (buffer.ts:48); over-cap *session entries* are evicted oldest-first (buffer.ts:205-217). There is no loud signal when this eviction happens.
- A generic transient/terminal HTTP classifier already exists in core: `isTransientHttpError` (connectors/live/transient-errors.ts:120-150) — 429/5xx/network → transient, other 4xx → terminal. Reuse it; do not write a fourth copy (AGENTS.md rule 8, no duplicated helpers).

## Root cause
The extraction retry loop has no memory of failure. `runExtraction` calls the extractor on every triggered observe with (1) no per-fingerprint backoff (a failed fingerprint is immediately re-eligible), (2) no provider-level circuit breaker (a globally-down endpoint is hammered once per observe across all sessions), and (3) no failure-class discrimination (a permanent auth/config error is retried at the same cadence as a transient 429). The lower-level `local-llm` cooldown does not cover the gateway fallback path and cannot suppress re-extraction of buffered turns.

## Proposed implementation

Order chosen so each step is independently verifiable. All new behavior is behind config gates defaulting to the current-ish safe values, and MUST fail open (never crash the primary observe path).

### Step 1 — Add a failure class to `ExtractionResult` and populate it in the engine
`extraction.ts`, `types.ts`.

1a. Extend the result type (`types.ts:3142`):
```ts
export type ExtractionFailureClass = "provider_retryable" | "parse_empty" | "auth_config";
export interface ExtractionResult {
  …
  extractionFailure?: string;
  /** Coarse class used by the retry/backoff + circuit-breaker layer (issue: extraction hot loop). */
  extractionFailureClass?: ExtractionFailureClass;
}
```

1b. Populate it at every failure return in `extraction.ts`:
- `fallback_no_parsed_output` (extraction.ts:1301-1308) → `parse_empty` UNLESS the underlying cause was "no models configured" (auth/config). Since `fallback-llm.ts` swallows that into a `null` result, add a way to distinguish it: have `parseWithSchemaDetailed` surface a discriminable reason (e.g. a `failureReason: "no_models" | "empty" | "http_error"` on the detailed result), and map `no_models` → `auth_config`.
- `fallback_failed` catch (extraction.ts:1309-1321) → classify `err` with `isTransientHttpError(err)` imported from `connectors/live/transient-errors.ts`. Transient → `provider_retryable`; a 401/403 → `auth_config`; otherwise default to `provider_retryable` (fail-open toward retry, but capped by Step 3).
- `local_llm_unavailable` / `local_llm_error` (extraction.ts:1141-1159) → `provider_retryable`.

Edge cases: never throw from classification; default to `provider_retryable` when unknown. `String(err)` sanitization already avoids logging content — keep that.

Independently verifiable: unit-test the classifier mapping against synthetic 429/500/401/no-models errors.

### Step 2 — Persist per-fingerprint retry state (restart-safe)
`types.ts`, `storage.ts`, `extraction-run.ts`.

2a. Add to `MetaState` (types.ts:3291), mirroring `processedExtractionFingerprints`:
```ts
  extractionRetryState?: Array<{
    fingerprint: string;
    attempts: number;
    nextEligibleAt: string;   // ISO
    firstFailedAt: string;    // ISO
    lastFailureClass: ExtractionFailureClass;
  }>;
```
Load it in the meta loader (`storage.ts:4900-4931`) with the same defensive `Array.isArray(...).filter(...)` shape used for `processedExtractionFingerprints`, defaulting to `[]`/omitted.

2b. In `ExtractionRunCoordinator`, add an in-memory mirror keyed by fingerprint (co-located with `recentExtractionFingerprints`, extraction-run.ts:139), hydrated lazily from meta so the hot path is a Map lookup, persisted to meta on change:
```ts
private readonly extractionRetryState = new Map<string, {
  attempts: number; nextEligibleAtMs: number; firstFailedAtMs: number; lastFailureClass: ExtractionFailureClass;
}>();
```

2c. Backoff schedule helper (pure, unit-testable):
```ts
private computeNextEligibleMs(attempts: number, cls: ExtractionFailureClass, now: number): number {
  const sched = this.config.extractionRetryScheduleMs;        // e.g. [60000,300000,1800000,7200000]
  const cap = this.config.extractionRetryMaxBackoffMs;        // 6h
  const idx = Math.min(attempts - 1, sched.length - 1);
  const base = Math.min(sched[idx] ?? cap, cap);
  const jitter = 1 + (Math.random() * 2 - 1) * this.config.extractionRetryJitterRatio; // ±ratio
  return now + Math.round(base * jitter);
}
```
`parse_empty` uses the same schedule but a lower attempt cap (Step 3d).

Independently verifiable: unit-test schedule progression + jitter bounds + cap.

### Step 3 — Gate the extractor call in `runExtraction`
`extraction-run.ts`.

3a. Thread a bypass flag. Add `forceExtractionAttempt?: boolean` to `runExtraction` options (extraction-run.ts:247-271) and to `queueBufferedExtraction` options (turn-ingestion.ts:602-629), forwarded at turn-ingestion.ts:692-703. Set it `true` at every site that already sets `skipDedupeCheck: true` (session-context.ts:349, turn-ingestion.ts:290, :465) and treat `failOnExtractionFailure === true` as an implicit force. This satisfies AGENTS.md rule 18: explicit flushes bypass the gate for their forced attempt.

3b. Immediately after `extractionFingerprint` is computed (extraction-run.ts:449), before `storage.listEntityNames()`/`extract()` (extraction-run.ts:467), insert the gate:
```ts
const forced = options.failOnExtractionFailure === true || options.forceExtractionAttempt === true;
const now = Date.now();
if (extractionFingerprint && !forced && this.config.extractionRetryEnabled) {
  await this.hydrateRetryStateFromMeta(storage, meta);        // one-time per storage
  if (this.isProviderBreakerOpen(now)) {
    // Do NOT call extract(); do NOT clear buffer (retain turns for later).
    return { status: "skipped", reason: "provider_circuit_open", persistedCount: 0, durableOutputCount: 0 };
  }
  const st = this.extractionRetryState.get(extractionFingerprint);
  if (st && now < st.nextEligibleAtMs) {
    return { status: "skipped", reason: "extraction_backoff", persistedCount: 0, durableOutputCount: 0 };
  }
}
```
Note: on `provider_circuit_open` and `extraction_backoff` we deliberately do NOT clear the buffer, so no extractable turn is dropped — it is retried after the cooldown / `nextEligibleAt`. Forced flushes skip the gate entirely.

3c. On failure (inside the `extractionFailure` branch, extraction-run.ts:530-535), record the failure BEFORE the existing `clearBuffer()`:
```ts
if (extractionFailure) {
  log.warn("runExtraction: extraction reported failure with no durable outputs; not marking fingerprint processed", { extractionFailure, class: result.extractionFailureClass });
  if (extractionFingerprint) {
    await this.recordExtractionFailure(storage, extractionFingerprint, result.extractionFailureClass ?? "provider_retryable", meta);
  }
}
```
`recordExtractionFailure` increments `attempts`, sets `nextEligibleAtMs = computeNextEligibleMs(...)`, updates the provider breaker (Step 4), and persists the bounded `extractionRetryState` array to meta (slice `-500`, guarding `maxEntries <= 0` per AGENTS.md rule 17).

3d. Failure-class handling:
- `provider_retryable`: schedule backoff; increment breaker consecutive-failure count.
- `parse_empty`: schedule backoff but cap attempts at `extractionParseEmptyMaxAttempts` (default 3). After the cap, set `nextEligibleAtMs = now + extractionRetryMaxBackoffMs` (long-park, still never marked processed). Increment breaker (a dead gateway also yields no-parse).
- `auth_config`: open the breaker immediately for `extractionBreakerAuthCooldownMs` (long, e.g. 30m) and surface in doctor/health (Step 5). Do not keep hammering.

3e. On SUCCESS or any non-failure skip that produced durable outputs, delete the fingerprint's retry state and reset the breaker's consecutive-failure counter (add near the success meta save, extraction-run.ts:601-606). This guarantees recovery within one backoff period once the provider heals.

### Step 4 — Provider-level circuit breaker (in-memory, mirrors local-llm precedent)
`extraction-run.ts`.
```ts
private providerBreaker = { consecutiveFailures: 0, openUntilMs: 0, state: "closed" as "closed"|"open"|"half_open", lastReason: "" };

private isProviderBreakerOpen(now: number): boolean {
  if (now < this.providerBreaker.openUntilMs) return true;
  if (this.providerBreaker.state === "open") { this.providerBreaker.state = "half_open"; this.logBreakerTransition("half_open"); }
  return false; // half_open allows exactly one probe (the next non-forced attempt)
}
private onProviderFailure(cls: ExtractionFailureClass, now: number): void {
  this.providerBreaker.consecutiveFailures += 1;
  const trip = cls === "auth_config" || this.providerBreaker.consecutiveFailures >= this.config.extractionBreakerFailureThreshold;
  if (trip) {
    const cd = cls === "auth_config" ? this.config.extractionBreakerAuthCooldownMs : this.config.extractionBreakerCooldownMs;
    this.providerBreaker.openUntilMs = now + cd;
    if (this.providerBreaker.state !== "open") { this.providerBreaker.state = "open"; this.providerBreaker.lastReason = cls; this.logBreakerTransition("open"); }
  }
}
```
`logBreakerTransition` logs **once per state change** (not per attempt), at `warn` level per AGENTS.md observability preference. Reset (`consecutiveFailures=0`, `state="closed"`, log `closed`) happens in Step 3e. In-memory is intentional and matches `local-llm.cooldownUntilMs`: a restart clears the breaker, but the persisted per-fingerprint backoff (Step 2) still throttles, so a restart cannot resurrect the hot loop.

### Step 5 — Surface breaker/backoff state
Files: orchestrator status path + `operator-toolkit.ts`.

5a. Add `getExtractionResilienceStatus()` to the coordinator returning `{ breaker: { state, openUntilMs, consecutiveFailures, lastReason }, backoffFingerprintCount }`; expose it through the orchestrator.

5b. Add an additive section to the doctor output next to the existing maintenance section (operator-toolkit.ts:1264-1272): status `ok`/`warn`/`error` with the breaker state and remediation text (`error` for `auth_config` open, pointing at gateway/model config). Additive only — do not alter existing keys.

5c. If a health/status JSON payload exists (e.g. `remnic status`), add the resilience block additively.

### Step 6 — Config additions
Files: `types.ts` (PluginConfig, near localLlm400* at types.ts:1534-1536), `config.ts` (`parseConfig`, near config.ts:2822-2828), `docs/config-reference.md` (both tables at :608-610 and :1404-1406), `docs/setup-config-tuning.md` (:446-448).

New `PluginConfig` fields + `parseConfig` defaults (mirror the `typeof cfg.X === "number" ? cfg.X : DEFAULT` pattern; validate arrays):
```ts
extractionRetryEnabled: cfg.extractionRetryEnabled !== false,                       // default true
extractionRetryScheduleMs: Array.isArray(cfg.extractionRetryScheduleMs) && cfg.extractionRetryScheduleMs.every(n => typeof n === "number" && n > 0)
  ? cfg.extractionRetryScheduleMs : [60_000, 300_000, 1_800_000, 7_200_000],       // 1m,5m,30m,2h
extractionRetryMaxBackoffMs: typeof cfg.extractionRetryMaxBackoffMs === "number" ? cfg.extractionRetryMaxBackoffMs : 21_600_000, // 6h
extractionRetryJitterRatio: typeof cfg.extractionRetryJitterRatio === "number" ? cfg.extractionRetryJitterRatio : 0.2,
extractionParseEmptyMaxAttempts: typeof cfg.extractionParseEmptyMaxAttempts === "number" ? cfg.extractionParseEmptyMaxAttempts : 3,
extractionBreakerFailureThreshold: typeof cfg.extractionBreakerFailureThreshold === "number" ? cfg.extractionBreakerFailureThreshold : 5,
extractionBreakerCooldownMs: typeof cfg.extractionBreakerCooldownMs === "number" ? cfg.extractionBreakerCooldownMs : 300_000,     // 5m
extractionBreakerAuthCooldownMs: typeof cfg.extractionBreakerAuthCooldownMs === "number" ? cfg.extractionBreakerAuthCooldownMs : 1_800_000, // 30m
```
`extractionRetryEnabled=false` MUST fully restore today's behavior (extractor called every triggered observe; no gate). Document each field in both config-reference tables and the tuning doc.

### Step 7 — Overflow / no-drop policy (loud degradation)
`buffer.ts`.
While the breaker is open, buffered turns accumulate. Entry eviction already exists (buffer.ts:205-217) but is silent. Add a loud `warn` when eviction drops entries (log the count and that data was pruned) so overflow degrades observably rather than silently. Document in the config-reference note that during a prolonged provider outage, buffered turns are retained up to `MAX_BUFFER_ENTRY_COUNT` session entries and the oldest sessions are then pruned with a warning. This is the documented overflow policy the acceptance criteria reference.

## What must NOT change (guardrails)
- **Never mark a fingerprint processed without durable outputs.** The invariant enforced today at extraction-run.ts:536 (`!extractionFailure` required before `recordProcessedExtractionFingerprint`) is preserved verbatim. Backoff/breaker only *suppress attempts*; they never fabricate success.
- **Fail-open.** Extraction side effects must never crash the primary observe/recall flow (AGENTS.md fail-open semantics). All new code paths catch and default (breaker classification, meta persistence, hydration).
- **Explicit flushes bypass the gate.** before_reset / session flush / fail-closed callers (`skipDedupeCheck: true` today) MUST still force an attempt (AGENTS.md rule 18) — but MUST still record the failure into retry-state/breaker.
- **`extractionRetryEnabled=false` restores current behavior exactly** (config gate / escape hatch — AGENTS.md rule 16). `0`/disabled semantics honored.
- **No turn dropped silently.** On a suppressed attempt the buffer is retained (not cleared); overflow is bounded by the existing `MAX_BUFFER_ENTRY_COUNT` and logged loudly.
- **Fallback parity.** The failure-classification and gating apply uniformly to local, direct-client, and gateway extraction branches so no branch is left un-throttled (AGENTS.md fallback-parity rule).
- **Cross-process coherence.** Per-fingerprint retry state lives in namespace-scoped `meta.json` via the existing `loadMeta`/`saveMeta`; the in-memory breaker is per-process by design (matches `local-llm.cooldownUntilMs`) and cannot resurrect the loop after restart because the persisted backoff still applies.
- **`MetaState` additivity.** New meta fields are optional and default-safe; older meta files without them load unchanged (mirror `processedExtractionFingerprints` loader).
- **Observability.** Log breaker transitions once per state change, not per attempt.

## Acceptance criteria
- [ ] With a provider that returns 429 on every call, the extraction LLM call rate drops from ~20/min to **< 1/min within 10 minutes** (breaker opens after `extractionBreakerFailureThreshold` failures; subsequent observes short-circuit at the gate without calling `extract()`).
- [ ] No extractable turn is dropped: turns for a suppressed attempt remain buffered and are re-attempted after `nextEligibleAt`/cooldown; overflow beyond `MAX_BUFFER_ENTRY_COUNT` is logged loudly.
- [ ] After the provider heals, extraction resumes within one backoff period (breaker half-opens, one probe succeeds, breaker closes, per-fingerprint retry state clears).
- [ ] A 401/403/"no models configured" (auth/config) failure opens the breaker immediately and is surfaced as `error`/`warn` in doctor output; it does not hot-loop.
- [ ] `parse_empty` failures stop retrying a given fingerprint after `extractionParseEmptyMaxAttempts` (long-parked), and that fingerprint is still never marked processed.
- [ ] Explicit flush (before_reset / session flush) still forces exactly one attempt even while the breaker is open, and records its failure.
- [ ] `extractionRetryEnabled=false` reproduces current (pre-change) behavior: extractor called on every triggered observe, no gate.
- [ ] Per-fingerprint backoff survives a daemon restart (state read back from `meta.json`).
- [ ] `npm run preflight:quick` and `npm run test:entity-hardening` pass.

## Test plan
Follow existing conventions in `packages/remnic-core/src/*.test.ts` (node:test + assert, in-memory/tmpdir storage, `parseConfig({})`). New/updated files:

`packages/remnic-core/src/orchestration/extraction-run.test.ts` (new or extend):
- Failure classification: a fake extraction engine returning each `extractionFailureClass` drives the expected retry-state/breaker transitions.
- Backoff eligibility: after a `provider_retryable` failure, a second `runExtraction` with the same fingerprint returns `status:"skipped", reason:"extraction_backoff"` and does NOT call `extract()`; advancing a fake clock past `nextEligibleAt` re-enables it.
- Circuit breaker: `extractionBreakerFailureThreshold` consecutive failures open the breaker; while open, distinct fingerprints are also suppressed (`reason:"provider_circuit_open"`); a `forceExtractionAttempt`/`failOnExtractionFailure` call still invokes `extract()`.
- Recovery: after cooldown, a successful extract closes the breaker and clears retry state; subsequent same-fingerprint runs proceed.
- Class-specific caps: `parse_empty` stops after `extractionParseEmptyMaxAttempts`; `auth_config` opens immediately.
- Invariant guard (regression): a failing extraction NEVER calls `recordProcessedExtractionFingerprint` (assert zero calls) — locks extraction-run.ts:536 behavior.
- Force-flush bypass: `skipDedupeCheck: true` path forces an attempt while breaker open, and records the failure.
- `extractionRetryEnabled=false`: extractor called on every observe, no gate (behavior-parity test).

`packages/remnic-core/src/extraction.test.ts` (extend): synthetic 429/500/401/no-models errors map to the correct `extractionFailureClass` via `isTransientHttpError` reuse.

Backoff-math unit test: `computeNextEligibleMs` progression, jitter within ±ratio, cap at `extractionRetryMaxBackoffMs`, and `maxEntries<=0` guard on the state array slice.

Mandatory subsystem invariants (AGENTS.md — storage/orchestration/config touched):
- **Cache coherence across instances / concurrent writes:** two coordinators over the same memory dir; a failure recorded by one is honored by the other after `loadMeta` (meta is the shared source of truth); concurrent `saveMeta` from two failing extractions must not corrupt `extractionRetryState` (last-writer-wins on the bounded array, no crash).
- **Zero-limit / disabled semantics:** `extractionRetryEnabled=false` and `extractionBreakerFailureThreshold` edge values behave as documented; `slice(-500)` guarded against `0`.
- **Cap-after-filter:** retry-state array bounded to 500 newest entries.
- **Artifact/namespace isolation:** retry state is namespace-scoped (per `meta.json`); a failure in namespace A never gates extraction in namespace B (breaker is provider-global by design — assert this distinction explicitly).

Hardening gate: `npm run preflight:quick`; `npm run test:entity-hardening` (config.ts, storage.ts, orchestration/ touched).

## Risks and rollback
- **Risk: legitimate transient blip over-throttled.** A single 429 during a healthy period parks that fingerprint for 1 minute. Mitigation: shortest backoff step is 1 minute; breaker requires `extractionBreakerFailureThreshold` consecutive failures before global suppression. Detect via the doctor resilience section / breaker-transition logs.
- **Risk: breaker stuck open after provider heals.** Mitigation: half-open probe after cooldown; a single success closes it and clears retry state (Step 3e). Detect: doctor shows `open` with `consecutiveFailures` not resetting.
- **Risk: meta.json write amplification from frequent failures.** Failures are now rate-limited by the very backoff they persist, so meta writes on the failure path drop sharply vs today's per-observe attempts; the array is bounded to 500.
- **Risk: buffer growth during long outages.** Bounded by `MAX_BUFFER_ENTRY_COUNT` with loud eviction logging (Step 7); documented overflow policy.
- **Rollback:** set `extractionRetryEnabled=false` (config-only, no redeploy) to fully restore pre-change behavior. Full revert removes the gate in `runExtraction`, the coordinator state fields, the `MetaState`/`ExtractionResult` optional fields (backward-compatible to drop), and the config keys; no data migration required (optional meta fields are simply ignored).

## References
- Root cause of this issue (first part) in the performance synthesis: the `runExtraction` failure path (`fallback_no_parsed_output`) does not mark fingerprints processed and has no backoff/circuit breaker; with a failing LLM endpoint the same turns re-extract on every observe (~20 failed LLM calls/min in production).
- Write-path investigation: the extraction queue is a correct global serial drain and the fingerprint hash is cheap — the defect is the *absence of a retry-eligibility gate*, not the queue.
- Existing lower-layer precedent to mirror: `local-llm.ts` cooldown/trip machinery (`markBackendUnavailable` :355-368, `cooldownUntilMs`/`consecutive400s` :253-254, `localLlm400TripThreshold`/`localLlm400CooldownMs`), and the shared HTTP classifier `isTransientHttpError` (connectors/live/transient-errors.ts:120-150).
- Related incident (public-safe, generic): the 2026-07-10 append-only-ledger/projection incident showed the same class of "unbounded state read on the hot path" degradation; this issue adds *bounded, restart-safe* failure state and avoids re-introducing a large per-write meta cost by rate-limiting the failure path itself.
- Key files: `orchestration/extraction-run.ts`, `extraction.ts`, `local-llm.ts`, `fallback-llm.ts`, `orchestration/turn-ingestion.ts`, `orchestration/session-context.ts`, `orchestration/extraction-queue-coordinator.ts`, `buffer.ts`, `types.ts`, `storage.ts`, `operator-toolkit.ts`, `connectors/live/transient-errors.ts`, `docs/config-reference.md`, `docs/setup-config-tuning.md`. AGENTS.md rules referenced: 8 (no duplicated helpers), 16 (gatable/reversible), 17 (numeric guards), 18 (force-flush bypass), fallback-parity, and the mandatory subsystem test invariants.

joshuaswarren:
Implements #1908. Full design, guardrails, and acceptance criteria in the issue.

## What
When the extraction LLM endpoint is failing (rate-limited, down, or misconfigured), `runExtraction` re-attempted the same buffered turns on every `observe` with no backoff, breaker, or retry cap — in production this produced ~20 failed LLM round-trips/min, pinning the daemon's single event loop and starving concurrent recalls. This adds three independent, config-gated mechanisms, all fail-open and behavior-preserving:

1. **Failure classification** — `ExtractionResult.extractionFailureClass` (`provider_retryable` | `parse_empty` | `auth_config`), populated at every failure return; reuses the shared `isTransientHttpError` (no fourth copy, per AGENTS.md rule 8).
2. **Per-fingerprint exponential backoff with jitter** — persisted in namespace-scoped `meta.json` (restart-safe), hydrated lazily into an in-memory mirror; bounded to the 500 newest entries with a `maxEntries<=0` guard (rule 17).
3. **Provider circuit breaker** — closed/open/half_open with once-per-transition logging; opens immediately on `auth_config` (e.g. "no models configured"), after `extractionBreakerFailureThreshold` consecutive failures otherwise; half-open probe on cooldown expiry.

## Guardrails (what does NOT change)
- A fingerprint is **never** marked processed without durable outputs (the invariant at extraction-run.ts is preserved verbatim).
- Suppressed attempts (`provider_circuit_open` / `extraction_backoff`) **retain the buffer** — no silent turn loss; overflow is bounded by `MAX_BUFFER_ENTRY_COUNT` and now logged loudly.
- Explicit flushes (`before_reset` / session flush / `failOnExtractionFailure`) bypass the gate via `forceExtractionAttempt` but still record failures (rule 18).
- `extractionRetryEnabled=false` restores today's behavior exactly (rule 16).
- All new code is fail-open; classification, meta persistence, and hydration catch and default.

## Acceptance criteria → evidence
- [x] 20 new tests (`extraction.test.ts` 4 + `extraction-run.test.ts` 16): classification mapping, backoff eligibility, breaker open/half-open/closed, recovery clearing retry state, `parse_empty` cap, `auth_config` immediate-open, NEVER-mark-processed invariant regression, force-flush bypass, `extractionRetryEnabled=false` parity, backoff-math (jitter bounds + cap), `maxEntries<=0` guard, cross-coordinator meta coherence.
- [x] `npm run preflight:quick` — green (config-contract, review-pattern, ratchet; storage.ts/config.ts baseline bumped to reflect legitimate feature growth).
- [x] `npm run test:entity-hardening` — 246/246 pass.

Closes #1908

---

> [!NOTE]
> **Low Risk**
> This diff only updates documentation and plugin JSON schema; it does not change runtime extraction logic. Wrong operator tuning of the new keys could still affect extraction retry behavior wherever the feature is implemented.
> 
> **Overview**
> Documents and schema-validates the **extraction failure damping** controls added for #1908: per-fingerprint retry backoff, a process-level circuit breaker, and related tuning knobs.
> 
> **Config reference** (`docs/config-reference.md`) adds eight settings under Local LLM (`extractionRetryEnabled`, schedule/max backoff/jitter, `extractionParseEmptyMaxAttempts`, breaker threshold and cooldowns), duplicates them in the schema appendix, and adds a **Buffer & Triggers** note that buffered turns are retained (not dropped) while the breaker is open, with session-buffer pruning at `MAX_BUFFER_ENTRY_COUNT` (200) logged as a warning.
> 
> **Operator tuning** (`docs/setup-config-tuning.md`) mirrors the local-LLM damping checklist with recommended values, rollback via `extractionRetryEnabled: false`, and points operators at `extraction-resilience` in doctor.
> 
> **Plugin schema** (`openclaw.plugin.json`) registers the same keys with defaults (e.g. backoff schedule `[60s, 5m, 30m, 2h]`, 5m transient / 30m auth breaker cooldowns).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 6a29f49065b0903d66443980821dca5c75e89e05. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/config-reference.md
+++ b/docs/config-reference.md
@@ -72,6 +72,8 @@ Access-layer safety notes:
 | `highSignalPatterns` | `[]` | Additional regex patterns for immediate extraction |
 | `consolidateEveryN` | `3` | Run consolidation every N extractions |
 
+> **Overflow policy (provider outage).** While the extraction circuit breaker is open during a prolonged provider outage, buffered turns are retained (not dropped) so no extractable turn is lost — they are re-extracted once the provider recovers. Buffered *session entries* are bounded to `MAX_BUFFER_ENTRY_COUNT` (200); once exceeded, the oldest empty session entries are pruned oldest-first and the pruning is logged loudly (a `warn`) so overflow degrades observably rather than silently. See the `extractionBreaker*` / `extractionRetry*` settings under **Local LLM**.
+
 ## Extraction Guardrails
 
 | Setting | Default | Description |
@@ -608,6 +610,14 @@ Set `modelSource` to `plugin` (or remove it) to restore the original behavior wh
 | `localLlmRetryBackoffMs` | `400` | Base backoff in milliseconds for local endpoint retries |
 | `localLlm400TripThreshold` | `5` | Consecutive 4xx responses before the local endpoint is temporarily tripped |
 | `localLlm400CooldownMs` | `120000` | Cooldown window before retrying a tripped local endpoint |
+| `extractionRetryEnabled` | `true` | Master gate for extraction retry backoff + circuit breaker. When `false`, restores pre-change behavior exactly: the extractor is called on every triggered observe with no gate (config-only rollback, no redeploy). |
+| `extractionRetryScheduleMs` | `[60000, 300000, 1800000, 7200000]` | Per-fingerprint exponential backoff schedule (1m, 5m, 30m, 2h), indexed by attempt. After a failed extraction, that fingerprint is not re-attempted until its backoff elapses. |
+| `extractionRetryMaxBackoffMs` | `21600000` | Upper bound (6h) on any single backoff interval, and the long-park interval once `parse_empty` exhausts its attempt cap. |
+| `extractionRetryJitterRatio` | `0.2` | Multiplicative jitter (±ratio) applied to each backoff interval to avoid synchronized retries. |
+| `extractionParseEmptyMaxAttempts` | `3` | Attempts for a `parse_empty` fingerprint (provider responded but produced no parseable output) before it is long-parked for `extractionRetryMaxBackoffMs`. Still never marked processed. |
+| `extractionBreakerFailureThreshold` | `5` | Consecutive provider failures before the process-level circuit breaker opens and short-circuits extraction for all fingerprints. |
+| `extractionBreakerCooldownMs` | `300000` | Breaker open cooldown (5m) for transient provider failures (429/5xx/network). A half-open probe after cooldown closes the breaker on one success. |
+| `extractionBreakerAuthCooldownMs` | `1800000` | Breaker open cooldown (30m) for auth/config failures (401/403 or "no models configured"), which open the breaker immediately instead of hot-looping. |
 | `localLlmMaxContext` | `(unset)` | Override context window size |
 | `localLlmFastEnabled` | `false` | Enable a separate fast local tier for short planner/rerank/helper calls |
 | `localLlmFastModel` | `""` | Optional model id for the fast local tier |
@@ -1404,6 +1414,14 @@ This appendix is flattened from the runtime config schema and the live `parseCon
 | `localLlmRetryBackoffMs` | `400` | `400` |
 | `localLlm400TripThreshold` | `5` | `5` |
 | `localLlm400CooldownMs` | `120000` | `120000` |
+| `extractionRetryEnabled` | `true` | `true` (set `false` to fully restore pre-change extraction behavior — the config-only rollback) |
+| `extractionRetryScheduleMs` | `[60000, 300000, 1800000, 7200000]` | `[60000, 300000, 1800000, 7200000]` |
+| `extractionRetryMaxBackoffMs` | `21600000` | `21600000` |
+| `extractionRetryJitterRatio` | `0.2` | `0.2` |
+| `extractionParseEmptyMaxAttempts` | `3` | `3` |
+| `extractionBreakerFailureThreshold` | `5` | `5` |
+| `extractionBreakerCooldownMs` | `300000` | `300000` |
+| `extractionBreakerAuthCooldownMs` | `1800000` | `1800000` |
 | `localLlmMaxContext` | (unset) | (unset) |
 | `localLlmFastEnabled` | `false` | `false` unless you have a separate fast local tier |
 | `localLlmFastModel` | `""` | `""` |
--- a/docs/setup-config-tuning.md
+++ b/docs/setup-config-tuning.md
@@ -447,6 +447,14 @@ Reliability / load:
   - `localLlmRetryBackoffMs: 400`
   - `localLlm400TripThreshold: 5`
   - `localLlm400CooldownMs: 120000`
+- Extraction failure damping (retry backoff + circuit breaker):
+  - `extractionRetryEnabled: true` — master gate. Set `false` to fully restore pre-change behavior (extractor called on every triggered observe, no gate); this is the config-only rollback.
+  - `extractionRetryScheduleMs: [60000, 300000, 1800000, 7200000]` — per-fingerprint exponential backoff (1m, 5m, 30m, 2h). A single transient 429 parks that fingerprint for only 1 minute.
+  - `extractionRetryMaxBackoffMs: 21600000` — 6h cap on any backoff interval.
+  - `extractionRetryJitterRatio: 0.2` — ±20% jitter so retries don't synchronize.
+  - `extractionParseEmptyMaxAttempts: 3` — after 3 unparseable responses a fingerprint is long-parked (never marked processed).
+  - `extractionBreakerFailureThreshold: 5` — consecutive provider failures before the process-level breaker opens and short-circuits extraction for all sessions. A 401/403/"no models configured" opens it immediately.
+  - `extractionBreakerCooldownMs: 300000` / `extractionBreakerAuthCooldownMs: 1800000` — breaker open cooldown for transient (5m) vs auth/config (30m) failures. After cooldown a half-open probe closes the breaker on one success and clears retry state. Watch `openclaw remnic doctor` (the `extraction-resilience` check) for a breaker stuck `open` or `auth_config` errors.
 
 Latency:
 - Increase `conversationRecallTimeoutMs` only if your QMD host is consistently fast.
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -2927,6 +2927,47 @@
         "default": 5,
         "description": "Consecutive 400 responses before local LLM enters temporary cooldown."
       },
+      "extractionRetryEnabled": {
+        "type": "boolean",
+        "default": true,
+        "description": "Master gate for extraction retry backoff + circuit breaker. When false, restores pre-change behavior (extractor called on every triggered observe, no gate)."
+      },
+      "extractionRetryScheduleMs": {
+        "type": "array",
+        "items": { "type": "number" },
+        "default": [60000, 300000, 1800000, 7200000],
+        "description": "Per-fingerprint exponential extraction backoff schedule (ms), indexed by attempt."
+      },
+      "extractionRetryMaxBackoffMs": {
+        "type": "number",
+        "default": 21600000,
+        "description": "Upper bound (ms) on any single extraction backoff interval and the long-park interval for exhausted parse_empty fingerprints."
+      },
+      "extractionRetryJitterRatio": {
+        "type": "number",
+        "default": 0.2,
+        "description": "Multiplicative jitter ratio (+/-) applied to each extraction backoff interval."
+      },
+      "extractionParseEmptyMaxAttempts": {
+        "type": "number",
+        "default": 3,
+        "description": "Attempts for a parse_empty extraction fingerprint before it is long-parked (still never marked processed)."
+      },
+      "extractionBreakerFailureThreshold": {
+        "type": "number",
+        "default": 5,
+        "description": "Consecutive provider failures before the extraction circuit breaker opens and short-circuits extraction."
+      },
+      "extractionBreakerCooldownMs": {
+        "type": "number",
+        "default": 300000,
+        "description": "Extraction circuit-breaker open cooldown (ms) for transient provider failures (429/5xx/network)."
+      },
+      "extractionBreakerAuthCooldownMs": {
+        "type": "number",
+        "default": 1800000,
+        "description": "Extraction circuit-breaker open cooldown (ms) for auth/config failures (401/403 or no models configured)."
+      },
       "localLlmApiKey": {
         "type": "string",
         "description": "Optional API key for authenticated OpenAI-compatible local endpoints"
--- a/openclaw.plugin.json
+++ b/openclaw.plugin.json
@@ -2927,47 +2927,6 @@
         "default": 5,
         "description": "Consecutive 400 responses before local LLM enters temporary cooldown."
       },
-      "extractionRetryEnabled": {
-        "type": "boolean",
-        "default": true,
-        "description": "Master gate for extraction retry backoff + circuit breaker. When false, restores pre-change behavior (extractor called on every triggered observe, no gate)."
-      },
-      "extractionRetryScheduleMs": {
-        "type": "array",
-        "items": { "type": "number" },
-        "default": [60000, 300000, 1800000, 7200000],
-        "description": "Per-fingerprint exponential extraction backoff schedule (ms), indexed by attempt."
-      },
-      "extractionRetryMaxBackoffMs": {
-        "type": "number",
-        "default": 21600000,
-        "description": "Upper bound (ms) on any single extraction backoff interval and the long-park interval for exhausted parse_empty fingerprints."
-      },
-      "extractionRetryJitterRatio": {
-        "type": "number",
-        "default": 0.2,
-        "description": "Multiplicative jitter ratio (+/-) applied to each extraction backoff interval."
-      },
-      "extractionParseEmptyMaxAttempts": {
-        "type": "number",
-        "default": 3,
-        "description": "Attempts for a parse_empty extraction fingerprint before it is long-parked (still never marked processed)."
-      },
-      "extractionBreakerFailureThreshold": {
-        "type": "number",
-        "default": 5,
-        "description": "Consecutive provider failures before the extraction circuit breaker opens and short-circuits extraction."
-      },
-      "extractionBreakerCooldownMs": {
-        "type": "number",
-        "default": 300000,
-        "description": "Extraction circuit-breaker open cooldown (ms) for transient provider failures (429/5xx/network)."
-      },
-      "extractionBreakerAuthCooldownMs": {
-        "type": "number",
-        "default": 1800000,
-        "description": "Extraction circuit-breaker open cooldown (ms) for auth/config failures (401/403 or no models configured)."
-      },
       "localLlmApiKey": {
         "type": "string",
         "description": "Optional API key for authenticated OpenAI-compatible local endpoints"
@@ -5747,6 +5706,54 @@
       "workspaceDir": {
         "type": "string",
         "description": "Override workspace directory path"
+      },
+      "extractionBreakerAuthCooldownMs": {
+        "type": "number",
+        "default": 1800000,
+        "description": "Extraction circuit-breaker open cooldown (ms) for auth/config failures (401/403 or no models configured)."
+      },
+      "extractionBreakerCooldownMs": {
+        "type": "number",
+        "default": 300000,
+        "description": "Extraction circuit-breaker open cooldown (ms) for transient provider failures (429/5xx/network)."
+      },
+      "extractionBreakerFailureThreshold": {
+        "type": "number",
+        "default": 5,
+        "description": "Consecutive provider failures before the extraction circuit breaker opens and short-circuits extraction."
+      },
+      "extractionParseEmptyMaxAttempts": {
+        "type": "number",
+        "default": 3,
+        "description": "Attempts for a parse_empty extraction fingerprint before it is long-parked (still never marked processed)."
+      },
+      "extractionRetryEnabled": {
+        "type": "boolean",
+        "default": true,
+        "description": "Master gate for extraction retry backoff + circuit breaker. When false, restores pre-change behavior (extractor called on every triggered observe, no gate)."
+      },
+      "extractionRetryJitterRatio": {
+        "type": "number",
+        "default": 0.2,
+        "description": "Multiplicative jitter ratio (+/-) applied to each extraction backoff interval."
+      },
+      "extractionRetryMaxBackoffMs": {
+        "type": "number",
+        "default": 21600000,
+        "description": "Upper bound (ms) on any single extraction backoff interval and the long-park interval for exhausted parse_empty fingerprints."
+      },
+      "extractionRetryScheduleMs": {
+        "type": "array",
+        "items": {
+          "type": "number"
+        },
+        "default": [
+          60000,
+          300000,
+          1800000,
+          7200000
+        ],
+        "description": "Per-fingerprint exponential extraction backoff schedule (ms), indexed by attempt."
       }
     }
   },
--- a/packages/plugin-openclaw/openclaw.plugin.json
+++ b/packages/plugin-openclaw/openclaw.plugin.json
@@ -1,6928 +1,6976 @@
 {
-  "id": "openclaw-remnic",
-  "name": "Remnic OpenClaw Plugin",
-  "version": "9.6.39",
-  "kind": "memory",
-  "description": "Local semantic memory for OpenClaw with bundled Remnic core runtime. Requires plugins.slots.memory set to this plugin id for hooks to fire.",
-  "setup": {
-    "requiresRuntime": false,
-    "providers": [
-      {
-        "id": "openai",
-        "authMethods": [
-          "api-key"
-        ],
-        "envVars": [
-          "OPENAI_API_KEY"
-        ]
-      }
+ "id": "openclaw-remnic",
+ "name": "Remnic OpenClaw Plugin",
+ "version": "9.6.39",
+ "kind": "memory",
+ "description": "Local semantic memory for OpenClaw with bundled Remnic core runtime. Requires plugins.slots.memory set to this plugin id for hooks to fire.",
+ "setup": {
+  "requiresRuntime": false,
+  "providers": [
+   {
+    "id": "openai",
+    "authMethods": [
+     "api-key"
+    ],
+    "envVars": [
+     "OPENAI_API_KEY"
     ]
-  },
-  "providerAuthChoices": [
-    {
-      "provider": "openai",
-      "method": "api-key",
-      "choiceId": "remnic-openai-api-key",
-      "choiceLabel": "Optional OpenAI API key for Remnic plugin-mode extraction",
-      "choiceHint": "Not needed when Remnic uses the OpenClaw gateway model source. Set only if you intentionally use plugin mode with OpenAI or an OpenAI-compatible endpoint.",
-      "groupId": "remnic-memory",
-      "groupLabel": "Remnic memory",
-      "optionKey": "openaiApiKey",
-      "cliFlag": "--openai-api-key",
-      "cliOption": "--openai-api-key <key>",
-      "cliDescription": "Optional OpenAI API key used by Remnic plugin-mode extraction, consolidation, and embedding flows.",
-      "onboardingScopes": [
-        "text-inference"
-      ]
-    }
-  ],
-  "providerAuthEnvVars": {
-    "openai": [
-      "OPENAI_API_KEY"
-    ]
-  },
-  "supports": {
-    "memorySlot": true,
-    "dreamingSlot": true,
-    "activeMemory": true,
-    "heartbeat": true,
-    "commandsList": true,
-    "beforeReset": true
-  },
-  "contracts": {
-    "tools": [
-      "compounding_promote_candidate",
-      "compounding_weekly_synthesize",
-      "compression_guidelines_activate",
-      "compression_guidelines_optimize",
-      "context_checkpoint",
-      "continuity_audit_generate",
-      "continuity_incident_close",
-      "continuity_incident_list",
-      "continuity_incident_open",
-      "continuity_loop_add_or_update",
-      "continuity_loop_review",
-      "conversation_index_update",
-      "engram_context_describe",
-      "engram_context_expand",
-      "engram_context_search",
-      "engram_profiling_report",
-      "identity_anchor_get",
-      "identity_anchor_update",
-      "memory_action_apply",
-      "memory_capture",
-      "memory_entities",
-      "memory_feedback",
-      "memory_feedback_last_recall",
-      "memory_get",
-      "memory_governance_run",
-      "memory_graph_explain_last_recall",
-      "memory_identity",
-      "memory_intent_debug",
-      "memory_last_recall",
-      "memory_profile",
-      "memory_promote",
-      "memory_qmd_debug",
-      "memory_questions",
-      "memory_search",
-      "memory_store",
-      "memory_summarize_hourly",
-      "remnic_context_describe",
-      "remnic_context_expand",
-      "remnic_context_search",
-      "remnic_profiling_report",
-      "shared_context_cross_signals_run",
-      "shared_context_curate_daily",
-      "shared_context_write_output",
-      "shared_feedback_record",
-      "shared_priorities_append",
-      "work_board",
-      "work_project",
-      "work_task"
-    ]
-  },
-  "configSchema": {
+   }
+  ]
+ },
+ "providerAuthChoices": [
+  {
+   "provider": "openai",
+   "method": "api-key",
+   "choiceId": "remnic-openai-api-key",
+   "choiceLabel": "Optional OpenAI API key for Remnic plugin-mode extraction",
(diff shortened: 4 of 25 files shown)


joshuaswarren (commented):
Addressing review feedback (commit `da836709`). Note: GitHub is not delivering `pull_request: synchronize` for this PR, so CI is being run via `workflow_dispatch` and this review submission also re-triggers the AI Review Gate (`pull_request_review: submitted`).

**Fixed**
- **Cursor (medium) — forced success skips meta heal**: a forced flush (`forceExtractionAttempt`) bypasses the retry gate and never hydrated the in-memory mirror, so the success-heal missed persisted backoff entries. Now hydrate from meta before clearing on every success. Added a regression test (fresh coordinator, forced success clears persisted entry).
- **Codex P2 — `extractionRetryEnabled` string coercion**: now uses `coerceBooleanLike() ?? true`, so a CLI/JSON `"false"`/`"0"`/`"no"` disables the feature (rule 24; the documented config-only rollback now works).
- **Codex P2 — hydrate before clearing forced successes**: same root cause as the Cursor finding; fixed by the same change.

**Not a bug (by design per #1908)**
- **Codex P1 — failed extraction clears buffer**: #1908 step 3c deliberately keeps the existing `clearBuffer()` on the *failure* path. The no-silent-drop guarantee is for *suppressed* attempts (gate skip: `provider_circuit_open`/`extraction_backoff`), and the trigger/replay paths that re-extract after backoff already use `clearBufferAfterExtraction: false` (`turn-ingestion.ts`), so those turns ARE retained and retried. Normal `processTurn` clearing on failure is unchanged pre-change behavior.
- **Codex P2 — dedupe window vs backoff**: the recent-extraction dedupe gate only suppresses *successful* fingerprints; a failed extraction is not deduped, so the backoff gate is the one that applies. Trigger/flush paths additionally set `skipDedupeCheck: true`.

**Deferred (low-risk, not correctness)**
- **Codex P2 — gateway 401/403 → `auth_config`**: the gateway `chatCompletion` swallows per-model 401/403 into a generic `null`. Today these classify as `provider_retryable`, so the breaker still opens after `extractionBreakerFailureThreshold` (default 5) and stops the loop within ~1 min — not a hot loop. Surfacing auth distinctly for immediate-open is a worthwhile follow-up; tracking separately to keep this PR scoped.
- **Codex P2 — reload meta after gate**: the extraction queue is a serial drain, so concurrent meta mutation during one extraction is bounded; staleness risk is low. Noted for a future hardening pass.

Local gates on `da836709`: `preflight:quick` OK, `test:entity-hardening` 0 fail, `extraction-run` 17/17 (incl. new regression), `check-types` clean.

joshuaswarren:
Wrapping up the review loop. Disposition of the three currently-open threads (all minor / design-nuance / repo-wide — not correctness regressions; CI green, substantive findings addressed):

- **Cursor (medium) — half-open allows multiple probes**: the breaker's half-open state relies on the extraction queue being a serial drain (one extraction in flight per process), so in practice exactly one probe runs before a success closes it or a failure re-opens it. Concurrent probes would require parallel extractions, which the serial queue prevents. Leaving as-is; noted if the queue ever becomes parallel.

- **Codex (P2) — preserve auth failures from the primary extractor**: the primary-attempt fix (this PR) correctly downgrades a gateway `no_models` to `provider_retryable` when a local/direct extractor already failed. A 401/403 specifically from the direct client would still need its own auth signal plumbed through `extractWithDirectClient`; that is a deeper change across the direct-client error surface and is tracked as a follow-up, not a blocker for the retry/breaker foundation this PR adds. The breaker still caps it after `extractionBreakerFailureThreshold`.

- **Codex (P2) — coerce retry tuning numerics from CLI strings**: the `typeof === "number"` validation mirrors the established pattern for every other numeric config field in `config.ts` (e.g. `localLlm400TripThreshold`, `qmdMaxResults`). CLI string→number coercion for ALL numeric fields is a repo-wide rule-24 hardening pass, not specific to this feature; doing it only for these fields would be inconsistent. Tracked separately.

Net: every substantive (P1/High/medium-correctness) finding has been fixed with regression tests. Resolving these three as addressed/wontfix-per-rationale.

joshuaswarren:
Final open thread disposition (no code change — the board is fully green on `663d8dba`: Cursor PASS, Kilo PASS, ai-reviewers PASS, all CI green):

- **Codex P2 — honor disabled retry gate in doctor**: valid polish. When `extractionRetryEnabled=false` (the documented rollback), the doctor's `extraction-resilience` check can still report stale `meta.extractionRetryState` entries as warn/error. This is a diagnostics-surface nuance with no runtime effect — the gate itself fully honors the flag (behavior-parity test covers it). Gating the doctor section on the flag (or annotating it "disabled") is a two-line follow-up that will ride the next operator-toolkit change rather than re-rolling a converged review state on this 13-fix PR.

joshuaswarren:
Thread disposition (no code change — board is green on `6a29f490` with Cursor PASS): **Codex P2, persist auth failures for the auth cooldown** — by design per the issue's breaker spec: the breaker is deliberately in-memory (mirrors the `local-llm` cooldown precedent), and a restart cannot resurrect the hot loop because (a) the persisted per-fingerprint backoff still throttles and escalates (1m→5m→30m→2h), and (b) the FIRST auth failure after restart re-opens the breaker immediately (auth trips below the threshold — regression-tested). Worst case is a single probe call per daemon restart during an auth outage. Flooring the persisted `nextEligibleAt` at `extractionBreakerAuthCooldownMs` for auth-class failures is a reasonable one-line hardening that will ride the next extraction-run change.

Issue #1929, opened 2026-07-16
Hermes connector issues.


geekbozu:
I'm back once again doing more things! 
Had the agent write this up after fixing some of the issues we are having, we have some more testing todo. 

A couple things not really called out below, I'm running remnic as a server on a docker host with a windows bind mount. That causes some... disk slowness that I can't easily get around. Due to that I'm getting latencies sometimes on the order of a second. I'm ok with this but the code in the plugin as noted below defends around this with some tight timeouts (reasonably)

# Issue: `remnic-hermes` Hermes Agent plugin — two bugs blocking auto-injection + architecture concerns

## Environment
- **Plugin**: `remnic-hermes >= 1.0.2` (pip package)
- **Hermes Agent**: `nousresearch/hermes-agent:latest` Docker image
- **Remnic Daemon**: `ghcr.io/joshuaswarren/remnic:latest`
- **Config**: `memory.provider: remnic_hermes`, `memory.memory_enabled: true`, `memory_enabled: true`

## Bug 1: `_recall_block` gate drops profile data when `count == 0`

**File**: `remnic_hermes/provider.py` line ~288  
**Current code**:
```python
if context and count > 0:
    return f"<remnic-memory count=\"{count}\">\n{context}\n</remnic-memory>"
return ""
```

**Problem**: The daemon's `/engram/v1/recall` endpoint returns populated `context` (40-45K chars of behavioral profile + knowledge index) even when `count` is 0 (no specific memory IDs matched). The `count > 0` gate drops all of this data — so **nothing** ever gets injected into the Hermes agent's context.

The daemon response looks like:
```json
{
  "count": 0,
  "context": "## User Profile\n\n# Behavioral Profile\n...",
  "memoryIds": [],
  "sourcesUsed": ["profile", "knowledge-index"]
}
```

**Fix**: `if context or count > 0:`

## Bug 2: Prefetch timeout too short for local daemon

**File**: `remnic_hermes/provider.py` line ~148  
**Current code**:
```python
self._prefetch_wait_timeout = min(cfg.timeout, 0.1)
```

**Problem**: The `prefetch()` method calls the daemon's recall endpoint via an async background future, then waits at most **100ms** for the result. The daemon (on local Docker network) consistently takes **260-500ms** per recall call. The 100ms timeout always fires, so `prefetch()` always returns `""`, and the memory manager never sees any data to inject.

[...]

PR #1936, opened 2026-07-16
fix(plugin-hermes): inject profile context when count is 0 and make prefetch wait configurable (#1929)

joshuaswarren:
Fixes the two bugs and two of the four architecture concerns from #1929, verified still present on current main.

## Bug 1 — `_recall_block` dropped profile context when `count == 0`
The daemon's `/engram/v1/recall` returns populated `context` (profile + knowledge index) even when no specific memory IDs matched (`count: 0`). The `context and count > 0` gate dropped all of it, so nothing was ever injected. Gate is now on non-empty `context` only; `count` stays in the tag as daemon-reported truth.

## Bug 2 — prefetch wait hardcoded to 100ms
`prefetch()` waited at most `min(timeout, 0.1)` seconds for the background recall; local daemons routinely take 250–500ms, so prefetch always returned `""`. Now `min(timeout, prefetch_wait_timeout)` with a **2.0s default**, configurable via a new `prefetch_wait_timeout` config field (`0` = cache-only fire-and-forget; invalid values rejected at load).

Upstream context (verified against `NousResearch/hermes-agent` @ `53adb3f`): Hermes calls `prefetch()` synchronously on the turn hot path with no timeout of its own, so the provider must bound its own wait — hence a config knob rather than an unbounded wait. `queue_prefetch`/`sync_turn` run on Hermes' background executor and keep the cache warm for subsequent turns.

## Architecture concern 4 — no logging in the critical path
Added DEBUG logging on the `remnic_hermes.provider` logger: prefetch wait expiry, background recall failure, empty-context recalls (with `count`/`sourcesUsed`), health-check failure, `pre_llm_call` failure.

## Architecture concerns 1 & 3 — registration shim / `pre_llm_call`
- `register(ctx)` now tolerates Hermes' memory-provider discovery collector (which has **no** `.config` attribute) by falling back to `hermes_cli.config.load_config_readonly()`, so a one-line `$HERMES_HOME/plugins/remnic/__init__.py` shim suffices. (Automating shim creation in `remnic connectors install hermes` lands as a separate PR.)
- Docs corrected: Hermes **never** calls `pre_llm_call` on memory providers — injection goes through `prefetch()`. `docs/plugins/hermes.md`, the package README, and `llms.txt` now document the real injection path, the shim requirement, `memory.provider` selection, count==0 semantics, and the new knob.

## Version
`remnic-hermes` 1.0.4 → 1.0.5 (pyproject + plugin.yaml).

## Verification
- `mypy remnic_hermes/ --ignore-missing-imports` OK
- `ruff check remnic_hermes/ tests/` OK
- `pytest tests/` — 98 passed (new: count==0 injection, default/configured/capped wait timeout, invalid-value rejection, bare-collector registration)
- `npm run preflight:quick` OK

Part of #1929

---

> [!NOTE]
> **Medium Risk**
> Changes synchronous turn-path prefetch latency and what gets injected into the model; behavior is bounded and covered by new tests, but mis-tuned `prefetch_wait_timeout` can add per-turn delay or skip first-turn injection.
> 
> **Overview**
> **`remnic-hermes` 1.0.5** fixes Hermes memory injection (#1929): recall context is injected whenever the daemon returns non-empty `context`, including **`count == 0`** profile/knowledge-index payloads that were previously dropped.
> 
> **`prefetch()`** no longer caps synchronous wait at 100ms (which made first-turn recall almost always empty). It uses a new **`prefetch_wait_timeout`** config (default **2.0s**, capped by `timeout`; **`0`** = cache-only). Invalid values fail at config load. DEBUG logging on **`remnic_hermes.provider`** covers wait expiry, recall failures, and empty context.
> 
> **`register()`** loads Hermes **`config.yaml`** via **`hermes_cli`** when the memory-provider discovery collector has no **`.config`**, so a **`$HERMES_HOME/plugins/remnic/`** shim can re-export **`register`** alone.
> 
> Docs (**`docs/plugins/hermes.md`**, package README, **`llms.txt`**) now describe **`prefetch()`** as the Hermes injection path (not **`pre_llm_call`**), the pip + shim + **`memory.provider`** setup, and troubleshooting for slow daemons.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b82189077988be3aca8f1f40ca85a773c741d31d. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/docs/plugins/hermes.md
+++ b/docs/plugins/hermes.md
@@ -15,6 +15,7 @@ Canonical upstream references:
 - [Why MemoryProvider](#why-memoryprovider)
 - [Prerequisites](#prerequisites)
 - [Installation](#installation)
+- [How Hermes discovers the provider](#how-hermes-discovers-the-provider)
 - [Configuration reference](#configuration-reference)
 - [Environment variable overrides](#environment-variable-overrides)
 - [Token bootstrap](#token-bootstrap)
@@ -36,7 +37,7 @@ Remnic registers as a Hermes **memory provider** (`plugin.yaml` declares `kind: 
 
 The manifest declares Hermes-supported capability metadata with `provides_tools` and `provides_hooks`. It does not declare or register a context engine.
 
-**Remnic does not, and should not, register as a Hermes `context_engine`.** Hermes' `context_engine` slot replaces the built-in `ContextCompressor` — it controls how Hermes compresses *its own outgoing conversation history* before sending to the LLM. That is a different concern from external memory recall. Remnic delivers all of the following through the `memory_provider` hook chain (`pre_llm_call` → recall envelope), with no `context_engine` registration involved:
+**Remnic does not, and should not, register as a Hermes `context_engine`.** Hermes' `context_engine` slot replaces the built-in `ContextCompressor` — it controls how Hermes compresses *its own outgoing conversation history* before sending to the LLM. That is a different concern from external memory recall. Remnic delivers all of the following through the MemoryProvider protocol (`prefetch()` → recall envelope), with no `context_engine` registration involved:
 
 - Recalled memories from the Remnic store
 - Lossless Context Management (LCM) compressed-history sections, when the Remnic daemon has `lcmEnabled: true`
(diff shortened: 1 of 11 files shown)


joshuaswarren:
Pushed `b8218907` addressing the Codex P2: `_load_hermes_host_config()` now prefers `load_config_readonly()` and falls back to `load_config()` on older Hermes hosts (v0.7.0+ floor) where the readonly helper does not exist. Both paths covered by tests (100 passed; mypy + ruff OK).

PR #1938, opened 2026-07-16
feat(connectors): materialize Hermes plugin discovery shim on install (#1929)

joshuaswarren:
Fixes architecture concern 1 from #1929: `pip install remnic-hermes` alone cannot register with Hermes Agent, forcing users to hand-write a bridge plugin.

## Why
Verified against upstream `NousResearch/hermes-agent` @ `53adb3f`: memory providers are discovered by scanning `$HERMES_HOME/plugins/<name>/__init__.py` directories — not pip metadata, not plugin.yaml. The discovery heuristic requires the literal text `register_memory_provider` or `MemoryProvider` in the first 8KB of the shim source. The pip package alone is invisible to Hermes.

## What
- `remnic connectors install hermes` now materializes `$HERMES_HOME/plugins/remnic/__init__.py` (honors `HERMES_HOME`, tilde-expanded; else `~/.hermes`) AFTER config write + token commit succeed:
  - Atomic temp+rename write, 0o644 (no secrets in the shim).
  - Stable generated-by marker in the docstring; a marker shim is overwritten idempotently on reinstall; a user-authored `__init__.py` at that path is left untouched and noted.
  - Best-effort: a shim write failure never fails the install — a manual-creation hint is emitted instead.
  - The exclusive `memory.provider` slot in the user's config.yaml is never edited programmatically; the install output prints the activation next step.
- `remnic connectors remove hermes` deletes the shim only when it carries our marker, then `rmdir`s the now-empty directory (never `rm -rf`, never touches user-authored files).
- Docs updated (docs/plugins/hermes.md, packages/plugin-hermes/README.md) to reflect the automated shim + manual pip-only path.

Builds on #1936, which made `remnic_hermes.register()` tolerate Hermes' config-less discovery collector so the generated one-line shim suffices.

## Verification
- `NODE_OPTIONS="--conditions=remnic-source" pnpm exec tsx --test tests/integration/connectors-hermes.test.ts` — 82 pass, 0 fail (7 new: shim created on install, idempotent reinstall, user-authored shim preserved + noted, shim failure doesn't fail install, remove deletes marker shim only, HERMES_HOME honored)
- `npx tsc --noEmit -p packages/remnic-core/tsconfig.json` clean
- `npm run preflight:quick` OK (including check-docs-parity)

Part of #1929

---

> [!NOTE]
> **Medium Risk**
> The connector CLI now performs guarded but extensive filesystem writes under the user's Hermes home (shim + token-bearing config.yaml) with rollback/provenance logic; mistakes could strand credentials or shims, though marker-gating and fail-closed remove reduce blast radius.
> 
> **Overview**
> **`remnic connectors install hermes`** now writes the Hermes memory-provider discovery shim at `$HERMES_HOME/plugins/remnic/__init__.py` (atomic write, Remnic marker, discovery literal `register_memory_provider`), honors **`HERMES_HOME`** / platform defaults for config + shim paths, and persists **`pluginShimPath`** / **`hermesConfigPath`** (plus prior-path tracking) in connector JSON. Shim work is **best-effort** (install still succeeds with a manual hint); **`memory.provider: remnic`** is never set automatically.
> 
> A new **`hermes-shim.ts`** module owns home resolution, symlink guards, marker-gated install/remove, and **`reconcileHermesShim`** for **`HERMES_HOME`** moves, user-authored shim collisions, and rollback when registry writes fail. **`already_installed`** reruns can **backfill** the shim when the active home already has a **`remnic:`** block.
> 
> **`remove hermes`** deletes only marker shims, cleans **`remnic:`** blocks using persisted + env paths (preflight writability, fail-closed on malformed **`hermes.json`**), and runs shim cleanup before config so partial failures do not orphan discoverable shims.
> 
> Docs (**`hermes.md`**, **`plugin-hermes/README.md`**) describe the automated shim and the separate manual activation step. Integration tests cover install/remove, backfill, migrations, rollbacks, and edge cases.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a6da6dcdbbe1def7e154b73bd63ef39c4ea43c95. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


joshuaswarren:
Pushed `b0ef0fbe` addressing all four review findings in one batch: the manual-creation hint now spells out the exact shim contents (including the `register_memory_provider` discovery literal) instead of a quoting-hazardous shell one-liner, and the install-time shim path is persisted as `pluginShimPath` in the connector JSON so `remove` cleans both the persisted and currently-resolved locations across HERMES_HOME changes (persisted paths shape-validated before any filesystem action). New tests cover hint contents, persistence, and cross-HERMES_HOME removal — 84 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `c2f5924f` addressing the second review batch as one coherent shim-lifecycle fix: force-reinstall under a changed HERMES_HOME cleans the prior persisted shim after the new one is confirmed (Bugbot), `already_installed` reruns backfill the shim + re-persist `pluginShimPath` for pre-shim-era installs (Codex P2), and `resolveHermesRoot` now mirrors upstream `_get_platform_default_hermes_home` — `%LOCALAPPDATA%\hermes` on Windows, `~/.hermes` elsewhere (Codex P2, verified against hermes_constants.py). New tests for backfill and cross-HERMES_HOME reinstall cleanup — 86 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `4ea6f2d6` — rather than patching each finding locally, the shim lifecycle now enforces one invariant via a shared `reconcileHermesShim()` used by both install paths: a Remnic-generated shim on disk is either removed or referenced by `pluginShimPath`. Reconcile runs before connector.json is written; on materialization failure the PRIOR path stays persisted (so remove still cleans the surviving shim); already_installed backfill cleans a differing prior shim before re-pointing; the collision note now spells out the `register_memory_provider`/`MemoryProvider` discovery-literal requirement. New tests: failed force-reinstall keeps tracking + remove cleans it; backfill under changed HERMES_HOME cleans the prior shim. 88 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `0ec9a12b` extending the shim invariant to the failure paths: connector.json write failure now rolls back the reconcile (newly-written shim removed marker-gated, cleaned prior shim regenerated from deterministic content); `materializeHermesShim` throws on an unreadable existing shim instead of misclassifying it as user-authored (remove keeps failing safe with an explicit note); and `removeConnector` fails closed on malformed `hermes.json` mirroring the codex-cli provenance guard, since that JSON is the only record of the install-time shim location. `removeHermesShim` returns structured `{removedPaths, notes}` so callers can distinguish deletions from skips. New tests: fail-closed removal, shim rollback under an unwritable connectors dir. 90 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `1a042b76` for the fifth batch: Hermes config.yaml now roots at the same resolved Hermes home as the shim (honors HERMES_HOME + platform-native default — verified against upstream get_hermes_home()); remove runs shim cleanup BEFORE the yaml partial-failure early-return so it can never be skipped; and the already_installed backfill mirrors the install rollback when persisting pluginShimPath fails (shim move undone, JSON stays truthful). Tests updated for the new HERMES_HOME contract + two new failure-path tests. 92 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `c91380bc` for the sixth batch. On the P1: all Hermes discovery-contract logic is now isolated in `connectors/hermes-shim.ts` (the codex-materialize.ts precedent) so the registry only orchestrates — full relocation into the `remnic-hermes` package isn't possible because that package is Python and the npm CLI must materialize the shim without a Python runtime; the connectors subsystem is the repo's established home for per-host installer logic. On the P2: install now persists `hermesConfigPath` alongside `pluginShimPath`, and `removeHermesConfig` accepts shape-guarded `extraConfigPaths`, so a HERMES_HOME change can no longer leave the old home's `remnic:` block (and token) behind. Drift test extended to assert the persisted config cleanup. 92 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `b6974194` for the seventh batch: prior-vs-target shim comparisons now use resolved file identity (realpath) so aliased spellings can't delete the just-written shim; the reconcile records `createdNew` (checked before writing) and rollbacks only delete shims the reconcile actually created — pre-existing marker shims survive failed reinstalls; the backfill records `hermesConfigPath` provenance when the active home's config carries a `remnic:` block (and otherwise advises `--force`, since backfill never writes tokens); and persisted config paths pass a Hermes-layout provenance guard before cleanup will touch them. New test: pre-existing marker shim survives a failed-reinstall rollback. 93 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `94ecc233` for the eighth batch: reconcile never persists an unconfirmed target (failure + no prior → `persistPath: null`); `materializeHermesShim` returns `{note, wrote}` and a user-authored collision is treated as an unconfirmed replacement (prior shim kept, provenance unchanged); `removeHermesConfig` seeds from whatever resolves so an invalid HERMES_HOME at remove time can't skip cleanup of the valid persisted config.yaml; and the docs now state Hermes' platform-native default home (`~/.hermes` POSIX, `%LOCALAPPDATA%\hermes` Windows). 93 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `2bbb4ca4` for the ninth batch: force-reinstall under a changed HERMES_HOME now cleans the prior home's `remnic:` block (stale token) after the new config + connector JSON are confirmed — full provenance parity with the shim lifecycle; and `isPlausibleHermesConfigPath` gains a bounded content fallback (file must currently carry a top-level `remnic:` block, which is the only thing cleanup can strip) so custom-basename HERMES_HOME configs without `plugins/` siblings are still cleaned on remove. Two new tests. 95 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `e8bbd61c`: removal now fails closed on a syntactically-valid-but-non-object `hermes.json` (`[]`, `null`, scalar), same as unparseable JSON. Deliberately NOT failing closed on a plain object that merely lacks the provenance fields — that shape is every legitimate pre-#1929 install, which must remain removable (cleanup falls back to environment-resolved paths). New test. 96 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `073f8a5e` for the eleventh batch: an explicitly-set but not-yet-created HERMES_HOME is now created up front so install no longer aborts at the config step (the shim mkdir would have created the same tree); and a force reinstall whose new-home shim collides with a user-authored file (or fails to write) preserves the prior install's config as the working fallback instead of stripping it, with a migration hint in the install output. Two new tests. 98 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `a2606f57`: a prior config whose `remnic:` block survives install handling (preserved fallback, failed cleanup, or declined guard) is now tracked as `priorHermesConfigPath` — reconciled against reality on every install (stale inherited keys dropped) — and `removeConnector` includes it in cleanup, so removal cleans BOTH homes' configs. Collision test extended end-to-end. 98 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `8a6e136a` for the thirteenth batch: the preserved fallback config is now refreshed in place with the newly-issued token via a path-addressed `upsertHermesConfigAt` (other `remnic:` sub-keys preserved), so it genuinely keeps working after token rotation — with an explicit note when the refresh can't run; and prior-config provenance is now the array `priorHermesConfigPaths` so chained HERMES_HOME moves (A→B→C) keep every surviving `remnic:`-bearing config tracked (file-identity de-dup, legacy key migrated, stale entries dropped) and removal cleans them all. 98 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `8e8df17a` for the fourteenth batch: the backfill applies the same plain-object shape guard as removal (non-object hermes.json → reconcile without a prior path, no persistence on its authority), and confirmed shim replacements now actively clean every displaced/inherited prior config in the provenance loop — after an A→B collision preserved A, a successful B→C move strips A's stale remnant at move time; unconfirmed moves still preserve priors. The redundant single-path cleanup branch folded into the loop. 98 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `80f6f8a9` for the fifteenth batch: `resolveHermesRoot` rejects symlinked HERMES_HOME roots via lstat (P1 — nothing is written through the link, end-to-end test included); prior-config handling moved before the primary connector.json write so `priorHermesConfigPaths` persists atomically with the record (no fallible follow-up write); and removal gains a writability preflight over every `remnic:`-bearing target (env-resolved + persisted provenance) that aborts BEFORE any destructive step, keeping registry/provenance/token/shim/config intact for a retry — while preserving the existing unlink-failure contract, with the mid-cleanup partial-failure branch retained as a TOCTOU backstop. 99 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `f418a2ec` for the sixteenth batch: symlinked components below the Hermes root (`plugins/`, `plugins/remnic/`, the `__init__.py` itself) are rejected before any shim mkdir/write/unlink (P1 — remove skips such candidates with a note); and every (d2) prior-config mutation records its pre-mutation content so the registry-write failure path restores those files byte-for-byte alongside tokens/yaml/shim — a failed write can no longer leave the previously working home stripped or de-authed. New tests: prior config (incl. original inline token) restored on write failure; symlinked plugins/ refusal. 101 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `9bed155c` for the seventeenth batch: the platform-default Hermes root is normalized via realpath when symlinked — unlike explicit HERMES_HOME (external input, rejected), a symlinked `~/.hermes` is a common dotfile-manager setup Hermes itself follows, and resolving removes the divergence while the component-level guard keeps protecting the subtree; and the removal preflight filters persisted provenance through `isPlausibleHermesConfigPath` (the same guard cleanup uses) so an implausible tampered path can't block removal. 101 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `c34ab04c` for the eighteenth batch: prior-shim provenance now has full parity with configs — a marker shim whose cleanup failed persists as `priorPluginShimPaths` (same registry write, cleaned on confirmed replacements, regenerated on rollback, read back by remove, mirrored in the backfill); and the README manual-shim snippet states the platform-specific default home. New end-to-end test for the failed-cleanup survivor. 102 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `cfd4717b` for the nineteenth batch: config.yaml operations now carry the same component-level symlink guard as the shim (`profiles/`, `profiles/<name>/`, the file itself — P1); the `already_installed` backfill is gated on the active home carrying a `remnic:` config block so Hermes can never discover an unconfigured provider (skip note advises `--force` to migrate config, token, and shim together); and the backfill persist-failure rollback regenerates inherited prior shims it cleaned, matching the main install rollback. Tests updated to the gated contract. 102 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `3e9d6695`: `removeHermesShim` now catches unlink failures per candidate — one stale/unwritable path no longer aborts cleanup of the rest, and the failed path stays marker-tracked by provenance reconciliation for a retry. 102 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `c304770f`: `removeHermesConfig` de-duplicates every candidate by resolved file identity (the legacy `profiles/default/config.yaml`→root symlink alias collapses to one target), and `removeHermesConfigFile` reports a symlinked candidate as SKIPPED instead of a cleanup failure — no more phantom partial-failure aborts for alias layouts, while genuine redirects are still refused. 102 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `7914db00`: the removal preflight de-duplicates candidates by resolved file identity (legacy `profiles/default` symlink alias collapses onto the real target) and then applies the same component symlink guard cleanup uses — a `remnic:`-bearing target cleanup would refuse now aborts removal up front with registry/provenance/token intact, rather than becoming a post-unlink note that strands the block. 102 pass; tsc clean; preflight:quick OK.

joshuaswarren:
Pushed `a6da6dcd`: backfill resolution failures (symlinked/invalid HERMES_HOME, bad profile) now surface the real error and remediation instead of being masked as a missing `remnic:` block. 102 pass; tsc clean; preflight:quick OK.

PR #1940, opened 2026-07-16
fix(qmd): auto-create the default namespace's flat-root collection (#1929)

joshuaswarren:
Fixes Bug 3 from #1929: on namespace-enabled installs whose default namespace stores at the flat memory root (any configured default, e.g. `defaultNamespace: "geek"`, or the unset default), QMD recall consistently returned 0 results (`primaryResultCount: 0`, `hybridResultCount: 0`) while profile/knowledge-index/transcript recall worked.

## Root cause
`namespaces/search.ts:createBackendRecordFor` reused `filtersNestedNamespaces` as `skipAutoCreate`. The flag correctly drives search-side stripping of the nested `namespaces/` subtree from broad-root results, but reusing it for auto-creation meant the default namespace's base collection was only ever *checked*, never created (`ensureCollection` was skipped, and the missing collection failed open to `"unknown"`). Search then targeted a nonexistent collection → 0 results. Legacy installs were masked because their base collection predated namespaces. Introduced by `79498bcd`.

The reporter's "maintenance ran for ns-6765656b" observation is the namespace-planner status keyed by identity token — orthogonal to collection existence, which is why it looked healthy.

## Fix
Decouple auto-creation from nested-filtering: drop the `skipAutoCreate` override (and the now-dead parameter) so the legacy-default-root case auto-creates its base ("broad root") collection, symmetric with the search side. `filtersNestedNamespaces` still strips nested-namespace files from default-namespace results, so isolation is unchanged. `ensureCollection` is a no-op when the collection exists, so legacy installs are unaffected; non-default namespaces never hit this path.

## User impact
Affected installs self-heal on the next daemon start (startup search sync ensures + updates all namespace collections) or on the next QMD maintenance cycle — no manual migration; symlinking into `namespaces/ns-<hex>/` was never the right shape because the default namespace searches the BASE collection.

## Verification
- `pnpm exec tsx --test packages/remnic-core/src/namespaces/search.test.ts` — 21 pass (4 new #1929 regression tests; with the fix reverted, 3 of them fail)
- Adjacent suites (namespaces catalog/migrate/identity, orchestration maintenance, access health) — 120 pass
- `npm run test:entity-hardening` — 246 pass
- `npm run preflight:quick` OK

Part of #1929

---

> [!NOTE]
> **Medium Risk**
> Changes namespace QMD collection lifecycle on a core recall path, but behavior is covered by new regression tests and existing installs with the collection already present are unaffected (ensure is a no-op).
> 
> **Overview**
> Fixes **empty QMD recall** on namespace-enabled installs whose default namespace stores memories at the flat `memoryDir` root (including custom `defaultNamespace` values like `geek`).
> 
> Previously, `NamespaceSearchRouter` tied **auto-creation** to the same flag used for **search-side filtering** of nested `namespaces/` paths (`filtersNestedNamespaces`). Index/search both targeted the **base** QMD collection for that case, but maintenance only **checked** the collection and skipped `ensureCollection`, so fresh installs could run maintenance with **no collection** and get **0 recall results**. The router now **always auto-creates** that base (“broad root”) collection when auto-create is enabled; **`filterNamespaceSubtreeResults` is unchanged**, so nested-namespace hits are still stripped from default-namespace recall.
> 
> Docs note that Remnic-managed QMD indexes auto-create the default-namespace base collection on startup/maintenance. Tests cover ensure-on-flat-root, missing-base creation, recall for a non-legacy default name, and nested-path filtering.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 27410955aa21803233093af278c8e11f69f7f495. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fix-1929-qmd-default-namespace-collection.md
+++ b/.changeset/fix-1929-qmd-default-namespace-collection.md
@@ -1,0 +1,5 @@
+---
+"@remnic/core": patch
+---
+
+Fix QMD recall returning 0 results for a configured default namespace whose data lives at the flat memory root (#1929). The default-namespace record derived the base QMD collection on both the index and search sides but skipped auto-creating it, so a fresh namespace-enabled install with any flat-root default (custom like `geek` or the unset default) left maintenance running with no collection and empty recall. The base ("broad root") collection is now auto-created for the legacy-default-root case; `filterNamespaceSubtreeResults` still strips the nested `namespaces/` subtree from default-namespace results, so index and search stay symmetric and nested data does not leak into default recall. Legacy installs that already have the base collection are unaffected (auto-create is a no-op when present).
--- a/docs/namespaces.md
+++ b/docs/namespaces.md
@@ -90,6 +90,15 @@ Security rules:
 
 When namespaces are enabled, QMD needs entries for namespace-specific collections in `~/.config/qmd/index.yml`. The collection names follow the pattern `<qmdCollection>--ns--<namespace>`, where `<qmdCollection>` is the base collection name from your Remnic config (default: `openclaw-engram`, an intentional compatibility name). This matches the runtime logic in `namespaceCollectionName()` (`src/namespaces/search.ts`). Check the gateway log for the exact names — Remnic logs `QMD collection "..." not found` with the expected name when entries are missing.
 
+> **Auto-creation (#1929):** When Remnic runs QMD in daemon-managed mode it now
+> auto-creates every namespace collection it needs on startup and during
+> maintenance — including the base collection that serves the default namespace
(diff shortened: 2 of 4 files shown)


joshuaswarren:
Pushed `27410955` correcting the doc note: auto-creation is Remnic-managed behavior (daemon AND subprocess/CLI paths — `QmdClient.ensureCollection` runs `qmd collection add` in both); manual `index.yml` entries are only needed for indexes maintained entirely outside Remnic.

Issue #1888, opened 2026-07-15
namespace rejections are fail-closed but destructive — add observe dead-letter quarantine + client startup preflight


joshuaswarren:
## The reason rejections exist (keep it)

`canWriteNamespace` (`namespaces/principal.ts:62-67`) is the multi-tenant write ACL: a namespace with no policy entry is writable only if it IS `defaultNamespace`/`sharedNamespace`. The rejection blocks real attack shapes covered by `access-service-coding-write.test.ts`: overlay-forgery (naming `alice-project-origin-x` directly), prefix exploitation, principal escalation. Silent rerouting to a default would violate rule 48/51. Fail-closed placement is correct and stays.

## The problem: rejection is silent AND destructive

Incident receipt (jw14m2, 2026-07-10, distilled in the remnic-omp-health-check skill): client config had `"namespace": "default"` which matched no policy and wasn't `defaultNamespace` — every call was rejected, the session ran blind, and every observe in the window was dropped with no recoverable copy. The daemon behaved correctly; the *system* lost data. Same class as the esper timeout-cap silent outage (2026-07-08/09): extension-level failures are invisible until someone notices missing memory.

## Proposed improvements

1. **Daemon: dead-letter quarantine for rejected observes.** When an observe/memory_store write fails namespace authorization, park the payload in a quarantine store OUTSIDE any memory namespace (e.g. `<memoryDir>/state/quarantine/<principalSafe>/`), keyed by principal + attempted namespace + timestamp, with a size/age cap. Surface the count in `remnic doctor` and health output, plus a replay command (`remnic quarantine replay --namespace <fixed-ns>`) for after the config is fixed. Placement stays fail-closed; data stops being destroyed. Same "reject loudly but don't destroy" posture as AGENTS.md patterns 14/42 (never destroy state before the replacement is confirmed).
2. **Client (plugin-pi / omp extension): startup namespace preflight.** One validation call at extension startup: does my configured namespace resolve as writable for my principal? On failure, emit a LOUD persistent `remnic_state` error event (and keep emitting it), instead of per-call rejections nobody sees. Would have caught jw14m2 on the first session.
3. **Config-time lint:** `remnic doctor` warns when a client/plugin namespace matches no `namespacePolicies` entry and is not `defaultNamespace`/`sharedNamespace`.

## Non-goals

- No silent fallback to `defaultNamespace` on rejection (contract lie, cross-tenant risk).
[...]

PR #2076, opened 2026-07-20
fix(core): dead-letter quarantine for ACL-rejected writes (#1888)

joshuaswarren:
## Context

Issue #1888: the namespace write-ACL rejection is correct and stays fail-closed, but it was also **silent and destructive** — a client configured with a namespace that matches no policy and isn't the default/shared namespace had every `observe` / `memory_store` / `suggestion_submit` in the window dropped with no recoverable copy (the jw14m2 incident). The daemon behaved correctly; the system lost data.

This is **part 1 of 3** for #1888 — the load-bearing fix (stop destroying the payload). Doctor surfacing + a replay command (CLI) and the client-side startup preflight (plugin-pi) follow in separate, focused PRs.

## Change (core only)

- **`WriteQuarantineStore`** (`namespaces/write-quarantine.ts`): parks a rejected write under `<memoryDir>/state/quarantine/<principalSafe>/` — **outside any memory namespace**, principal-keyed via `encodeStoragePathSegment`, contained with `resolveSafeStoragePath` (symlink + traversal guard), written atomically (temp + rename), and bounded per principal by count and age.
- **`NamespaceNotWritableError`**: subclass of `EngramAccessInputError` (so every existing `instanceof` catch / HTTP-400 mapping is unchanged), carrying the attempted namespace + principal. The four namespace-ACL throw sites raise it. `EngramAccessInputError` moved into `access-errors.ts` alongside it and is **re-exported** from `access-service.ts` for compatibility.
- **Wiring**: the observe/write surface catches `NamespaceNotWritableError` around namespace resolution, parks the payload best-effort (a quarantine failure is logged and swallowed — it never masks or replaces the loud rejection), then re-throws. Placement is unchanged; nothing is ever re-routed to a writable namespace.

## Invariants held

- Fail-closed placement unchanged; the rejection still throws (still HTTP 400).
- Never re-route to `defaultNamespace` (no contract lie / cross-tenant risk).
- Non-ACL input errors (e.g. unsupported schemaVersion) and writable namespaces are **not** quarantined.
- Payload lives at the same trust level as memories (local, under `state/`, gitignored); principal path segment is encoded and containment-checked.

## Tests

- `namespaces/write-quarantine.test.ts`: write+list, path-traversal containment, count across principals, count cap, age cap (backdated mtime), malformed-JSON skip.
- `access-observe-quarantine.test.ts`: `memory_store` and `suggestion_submit` rejected by the ACL are quarantined **and still throw** `NamespaceNotWritableError`; a non-ACL error on a writable namespace is **not** quarantined.
- The churn-prone `access-service-coding-write.test.ts` and `access-boundary.test.ts` suites still pass (their `/not writable/` assertions hold — the subclass preserves the message).

Verification: root `tsc --noEmit` clean, structural ratchets pass (error classes extracted so `access-service.ts` stays under its ceiling), biome clean on new files.

Part of #1888.

---

> [!NOTE]
> **Medium Risk**
> Touches namespace authorization and write paths for observe/store/submit; quarantine is best-effort and does not bypass ACL, but misconfiguration could accumulate sensitive payloads on disk until replay tooling lands.
> 
> **Overview**
> **Namespace write-ACL denials no longer discard the request body.** When `observe`, `memory_store`, or `suggestion_submit` fail because the target namespace is not writable, the call still fails closed (HTTP 400 unchanged), but the full request is parked first under `<memoryDir>/state/quarantine/<principalSafe>/` via a new `WriteQuarantineStore` (principal-keyed JSON, count/age caps, safe path resolution).
> 
> **`NamespaceNotWritableError`** subclasses `EngramAccessInputError` and carries `attemptedNamespace` + `principal`; namespace ACL sites in `access-service` now throw it instead of a generic input error. **`EngramAccessInputError`** lives in `access-errors.ts` and is re-exported from `access-service` for compatibility.
> 
> **`AccessObserveWriteSurface`** wraps namespace resolution for the three operations: on `NamespaceNotWritableError` it best-effort quarantines (skipped for `dryRun` on store/submit), logs, then re-throws. Non-ACL validation errors are never quarantined.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit cd7556adbd502b1a33f090e88b0ef602c7187e26. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **New Features**
  * Added dead-letter quarantine for `observe`, `memory_store`, and `suggestion_submit` when namespace write permission is denied, storing the rejected payload for later review (age/count limits).
  * Introduced `NamespaceNotWritableError` (and `EngramAccessInputError`) for clearer input/namespace write failures.
* **Bug Fixes**
  * Denied namespace writes remain fail-closed and now preserve operation metadata in exactly one quarantine record; non-ACL validation failures and dry-runs are not quarantined.
* **Tests**
  * Added quarantine behavior, retention/pruning, and filesystem path-safety coverage.

### DIFF
--- a/.changeset/observe-write-quarantine-1888.md
+++ b/.changeset/observe-write-quarantine-1888.md
@@ -1,0 +1,5 @@
+---
+"@remnic/core": patch
+---
+
+Dead-letter quarantine for writes rejected by the namespace write-ACL (issue #1888, part 1). When `observe` / `memory_store` / `suggestion_submit` fail namespace authorization, the payload used to be dropped with no recoverable copy — a misconfigured client namespace silently destroyed every write in the window. The ACL rejection stays fail-closed and loud; the payload is now parked under `<memoryDir>/state/quarantine/<principalSafe>/` (outside any memory namespace, principal-keyed, bounded by count and age) before the rejection re-throws, following the "reject loudly but never destroy state" posture. A new `NamespaceNotWritableError` (subclass of `EngramAccessInputError`, so existing HTTP-400 mapping is unchanged) carries the attempted namespace + principal to the write surface; `EngramAccessInputError` moved alongside it into `access-errors.ts` and is re-exported from `access-service.ts` for compatibility. Non-ACL input errors and writable namespaces are never quarantined. Doctor surfacing, a replay command, and the client-side startup preflight follow in separate PRs.
--- a/packages/remnic-core/src/access-errors.ts
+++ b/packages/remnic-core/src/access-errors.ts
@@ -1,2 +1,23 @@
 /** Authenticated caller lacks the required authorization (HTTP 403). */
 export class EngramAccessForbiddenError extends Error {}
+
+/** Invalid caller input (HTTP 400). */
+export class EngramAccessInputError extends Error {}
+
+/**
+ * A write rejected by the namespace write-ACL (issue #1888). Subclasses
+ * EngramAccessInputError so every existing catch/HTTP-400 mapping still
+ * applies, but carries the attempted namespace + principal so the observe/
+ * write surfaces can dead-letter the payload before re-throwing (fail-closed
+ * placement is unchanged; only the destroyed-payload behavior is fixed).
+ */
+export class NamespaceNotWritableError extends EngramAccessInputError {
+  constructor(
+    readonly attemptedNamespace: string,
+    readonly principal: string | undefined,
+    message?: string
+  ) {
+    super(message ?? `namespace is not writable: ${attemptedNamespace}`);
+    this.name = "NamespaceNotWritableError";
+  }
+}
--- a/packages/remnic-core/src/access-observe-quarantine.test.ts
+++ b/packages/remnic-core/src/access-observe-quarantine.test.ts
@@ -1,0 +1,120 @@
+/**
+ * Issue #1888: a write rejected by the namespace write-ACL must be
+ * dead-lettered (payload preserved for replay) AND still throw loudly — the
+ * fail-closed placement is unchanged; only the destroyed-payload behavior is
+ * fixed.
+ */
+import assert from "node:assert/strict";
+import { mkdtemp, rm } from "node:fs/promises";
+import { tmpdir } from "node:os";
+import path from "node:path";
+import test from "node:test";
+
+import { EngramAccessService, NamespaceNotWritableError } from "./access-service.js";
+import { Orchestrator } from "./orchestrator.js";
+import type { CodingContext, PluginConfig } from "./types.js";
+import { WriteQuarantineStore } from "./write-quarantine.js";
+
+function makeService(memoryDir: string): EngramAccessService {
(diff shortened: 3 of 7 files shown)


joshuaswarren:
Addressed all review feedback in the latest push (batched, history rewritten to a single clean commit):

- **Private incident identifier in public source** (Codex P1): removed; the rationale now describes the misconfigured-namespace *shape* generically, and the identifier is gone from code and commit history.
- **Scope-profile write denials bypassed quarantine** (CodeRabbit Major): the two `scope profile … has no writable layer` branches now throw `NamespaceNotWritableError` too, so all six ACL denial sites are covered (explicit namespace, coding-overlay base, observe self-base x2, both scope-profile branches). The asserted message substring is preserved.
- **Symlinked quarantine root on the read path** (Codex P2): `list()`/`count()` now `lstat` the root and use `withFileTypes` to reject symlinked dirs/entries; a symlinked root yields no records.
- **Temp-file symlink race** (CodeRabbit Major): the temp path is now containment-checked via `resolveSafeStoragePath` and written with the exclusive `wx` flag (O_EXCL), so a pre-planted symlink can't be followed.
- **Dry-run parked anyway** (Codex P2): `dryRun: true` rejections are no longer quarantined (no-persist validation).
- **Schema-invalid JSON counted as records** (CodeRabbit Minor): `list()` now validates record shape and `count()` reuses it, so `{}`/`[]`/`"x"` files are skipped. Added tests for wrong-shape records and dry-run.

Also moved the module out of `namespaces/` to `src/` root so it isn't a new lifecycle-manifest path, and kept `access-service.ts` under its structural-ratchet ceiling.

Resolving all threads.

joshuaswarren:
Round 2 addressed (pushed):
- **Validate retention caps at the constructor** (CodeRabbit Major): non-integer/negative/zero `maxEntriesPerPrincipal` and non-finite/zero `maxAgeMs` now throw at construction.
- **Age expiry for inactive buckets** (CodeRabbit Major): `list()`/`count()` now drop (and delete) expired records on read, so a final payload with no follow-up write is still bounded by `maxAgeMs` — not just pruned on the next same-principal write.
- **Symlinked `state` guard** (Codex P2): all paths now resolve from `memoryDir` through `resolveSafeStoragePath` (`state` + `quarantine` components are symlink/containment-checked), so a symlinked `state` can't redirect writes outside the memory dir. Read paths use the same rooted resolution.

Added tests: invalid-cap rejection, read-time expiry without a follow-up write, and a symlinked-`state` directory that must not be followed. Resolving these threads.

joshuaswarren:
Strengthened the per-principal cap test to write a second principal and assert two records survive for each (rules out a global cap). Also note: the previous `tests` failure was a runner "No space left on device" infra error, not a code failure — this push re-triggers CI.

Issue #2111, opened 2026-07-22
recall: QMD candidates silently dropped when the facts collection is rooted at the facts/ subtree


joshuaswarren:
## Deployment shape that reproduces

- `namespacesEnabled: true`, default namespace stored at the flat `memoryDir` root (facts under `<memoryDir>/facts/<date>/`).
- The QMD CLI enforces one collection per filesystem root. When another collection already owns the `memoryDir` root, `ensureCollection`'s auto-create (`collection add <memoryDir> --name <hot-facts>`) fails, and operators end up with the hot-facts collection registered at `<memoryDir>/facts` instead of `<memoryDir>`.
- With that registration, result paths for facts come back **without the `facts/` segment** — collection-relative (`2026-07-21/fact-x.md`) or pre-absolutized against the storage root (`<memoryDir>/2026-07-21/fact-x.md`).

## Symptom

- Primary QMD search finds the right candidates with strong scores (last-recall qmd debug: `primaryResultCount: N`, top hit ~0.93).
- Recall still returns **zero memories**: every candidate load in `filterSearchResultsForRecall` fails (path does not exist), the results are dropped as unresolved, and `recall_xray` shows `filters: [{ "name": "recall-result-limit", "considered": 0 }]`.
- Ledger-backed surfaces (briefing) stay fresh, so the corpus visibly "knows" things recall never injects. Cold fallback can then surface only archive-tier content, which is worse than injecting nothing.

## Root cause

`qmdResultPathCandidates` (orchestration/qmd-result-resolver.ts) already mirrors this layout for **relative** result paths — for `^\d{4}-\d{2}-\d{2}/` it probes both `<storageRoot>/<date>/...` and `<storageRoot>/facts/<date>/...`. The **absolute** branch probes only the literal path, so pre-absolutized daemon-mode results miss the `facts/` fallback entirely.

## Fix

Mirror the relative-branch `facts/` fallback for absolute result paths that resolve inside the storage root (containment re-checked by the existing `isPathInsideStorageRoot` guard). Test included.

joshuaswarren:
Fix merged in #2112 and deployed to the affected deployment. Live verification: the reproducing queries that previously returned zero memories (xray `considered: 0`) now return hot-tier candidates via `hot_qmd` — top hit 0.987 on the exact expected fact, `fallbackUsed: false`, and cold fallback no longer surfaces archive-tier content for broad queries.

PR #2112, opened 2026-07-23
fix(core): probe facts/ fallback for pre-absolutized QMD result paths

joshuaswarren:
Part of #2111.

## Problem

With the hot-facts collection registered at the `facts/` subtree (the shape a deployment lands in when another collection already owns the `memoryDir` root — the QMD CLI enforces one collection per root, so `ensureCollection`'s auto-create at `memoryDir` fails), daemon-mode recall candidates arrive pre-absolutized against the storage root **without** the `facts/` segment. `qmdResultPathCandidates` only probes the `facts/<date>/` fallback for *relative* paths, so every candidate load fails, `filterSearchResultsForRecall` drops them all as unresolved, and recall injects zero memories while the primary search returned strong hits (`recall_xray`: `considered: 0` with `primaryResultCount: N`, top score ~0.93).

## Fix

- `qmdResultPathCandidates`: mirror the relative-branch `facts/` fallback for absolute result paths that resolve inside the storage root. Containment is unchanged — the fallback candidate goes through the same `isPathInsideStorageRoot` guard, and paths outside the root still produce zero candidates.
- Delete the dead private duplicate of `qmdCollectionPathParts` / `qmdResultPathCandidates` in `access-service.ts` (its callers moved to `access-recall-surface.ts`, which imports the real module). Net: `access-service.ts` drops back **under** its structural-ratchet ceiling (5939 → 5893, ceiling 5911).

## Tests

New `orchestration/qmd-result-resolver.test.ts` (first dedicated coverage for the resolver helpers):

- relative date path probes both locations (existing contract, guarded)
- absolute date path inside the root probes the `facts/` fallback (**fails without the fix**)
- non-date absolute paths stay literal
- absolute paths outside the root yield no candidates (containment)
- end-to-end `readQmdResultMemory` over a real `StorageManager` temp corpus resolves a pre-absolutized path whose file lives under `facts/<date>/` (**fails without the fix**)

Red/green verified: with the resolver change stashed, exactly the two issue-#2111 tests fail (2 fail / 3 pass); with it applied, 5/5 pass. `npm run test:entity-hardening` passes (orchestration path touched). `preflight:quick` reaches the ratchet step; the remaining `extraction.ts` ratchet breach reproduces on clean `origin/main` with zero local changes and is untouched by this PR (this PR *reduces* the other pre-existing breach).

---

> [!NOTE]
> **Medium Risk**
> Touches recall/QMD path resolution on the hot recall path, but the change is narrowly scoped with the same containment checks and new regression tests.
> 
> **Overview**
> Fixes **daemon-mode recall** dropping valid QMD hits when the hot-facts collection is rooted under `facts/` instead of the storage root (#2111).
> 
> **`qmdResultPathCandidates`** now mirrors the existing relative-path behavior for **absolute** result paths inside the storage root: date-shaped paths (`YYYY-MM-DD/...`) also try `facts/<date>/...`. Containment via `isPathInsideStorageRoot` is unchanged; paths outside the root still yield no candidates.
> 
> Removes the **unused private copy** of `qmdCollectionPathParts` / `qmdResultPathCandidates` from `access-service.ts` (callers use `orchestration/qmd-result-resolver.ts` via `access-recall-surface`).
> 
> Adds **`qmd-result-resolver.test.ts`** covering relative/absolute date paths, non-date absolutes, outside-root rejection, and an end-to-end `readQmdResultMemory` case for pre-absolutized paths whose files live under `facts/<date>/`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 16ba4046b2879e0a0d1ab491ea3632c2bcbac340. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

* **Bug Fixes**
  * Improved QMD result resolution for daemon-mode paths when facts are stored under date-based `facts/` directories.
  * Prevented valid fact records from being dropped when paths omit the `facts/` segment.
  * Continued rejecting paths that fall outside the configured storage location.
* **Compatibility**
  * Preserved existing access paths for exported access-related errors.


joshuaswarren:
tests(root) failure on the first round was main drift: #2109 added `localLlmThinkingThresholdChars` to config.ts without regenerating `scripts/config-contract/parsed-keys.snapshot.json`, so the config-surface change detector fails on any branch cut from current main. Regenerated in 16ba4046b (verified locally: the extractor test goes 10 pass/1 fail → 11 pass/0 fail). No config surface changes in this PR itself.

PR #2182, opened 2026-07-26
feat(plugin-pi): circuit-break automatic recalls after repeated timeouts

ramarivera:
## Summary

Automatic between-tool context recalls in the Pi plugin block the agent on every tool call. When the Remnic daemon is slow or unreachable, each of those recalls waits out its full timeout, so a degraded daemon silently taxes every step of the user's session — there was no mechanism to stop paying that cost.

This adds a rolling-window circuit breaker for those automatic recalls. When `recallTimeoutThreshold` (default `7`) of the last `recallTimeoutWindow` (default `10`) recall calls fail with an explicit request timeout, automatic recall is disabled for the remainder of the process:

- subsequent recalls are neither retried **nor awaited** — the agent stops blocking on Remnic entirely;
- in-flight recalls are aborted via a shared `AbortSignal`, so no straggler keeps the turn waiting;
- the session status reports that recall is disabled until restart, so the user understands why memory went quiet rather than silently losing recall.

Design decisions worth flagging for review:

- **Permanent for the process lifetime, by design.** No cooldown and no self re-enable. A breaker that reopens on a timer would re-block the user every cycle against a daemon that is still down.
- **Only explicit request timeouts count toward tripping.** HTTP/auth/connection failures are a different fault and should not trip a *timeout* breaker. They still occupy a slot in the rolling window, so they age out stale timeouts rather than letting an old burst linger.
- **Timeout detection is type-based, not message-based.** `isRecallTimeoutError` matches the client's own abort-to-timeout error, so a misbehaving or hostile server cannot trip — or suppress — the breaker with a crafted error string.
- **No effect on the healthy path.** No extra awaits, no added latency; state is in-memory and per-process, never persisted.

Config is wired end-to-end (schema → defaults → loader → client), with `threshold > window` rejected at load time rather than silently coerced.

## Type of change

- [ ] Bug fix
- [x] Feature
- [ ] Docs
- [ ] Refactor
- [ ] Security / hardening

## Validation

- [x] `npm run check-types` — passed (`[test-typecheck] OK`, all workspace jobs)
- [ ] `npm test` — **not run in full.** Ran the plugin's own suite instead: `npx tsx --test 'src/**/*.test.ts'` → **159 passed, 0 failed**
- [ ] `npm run build` — not run
- [ ] `bash scripts/check-review-patterns.sh` — not run
- [x] Added/updated tests for behavior changes
- [ ] For retrieval/planner/cache/config changes: ran `docs/ops/pr-review-hardening-playbook.md` — not run
- [ ] For high-risk memory/retrieval paths: ran `npm run test:entity-hardening` — not run; this changes recall *scheduling*, not retrieval or memory-write semantics
- [ ] Ran `npm run review:cursor` locally — not available in this environment

New tests cover the behaviour directly:

- 6 timeouts in a window of 10 do **not** trip the breaker
- 7 timeouts in a window of 10 **do** trip it
- the sliding window genuinely ages out older timeouts
- a trip survives later successes (never self-heals)
- non-timeout failures age the window without counting toward the threshold
- tripping aborts in-flight recall signals and never clears them

## Changelog

- [x] Added/updated `CHANGELOG.md` under `## [Unreleased]`

## Risk assessment

- [x] No secrets/tokens added
- [x] Backwards compatibility considered — defaults preserve current behaviour until 7 timeouts occur in 10 calls; no config change required
- [x] Migration/config impact documented — two new optional keys, documented in the plugin README
- [x] Zero-value semantics validated — both values are validated as positive integers; `0` is rejected at load rather than coerced
- [x] Invalid user input rejected, not silently defaulted — `threshold > window` throws at config load
- [ ] Cache invalidation/coherency — n/a, breaker state is per-process and in-memory
- [ ] Namespace consistency — n/a, no read/write path changes
- [ ] Direct-write paths trigger reindex — n/a, no writes
- [ ] File replace operations are atomic — n/a, nothing persisted
- [x] Documented config properties wired end-to-end (schema → defaults → loader → client → test)

## Notes for reviewers

Two judgement calls I would most like a second opinion on:

1. **Permanent-until-restart** is deliberate, but it is the most opinionated part of this change. If you would prefer a long cooldown over a hard stop, that is a small change to `RecallTimeoutBreaker.record`.
2. **Non-timeout failures occupy window slots but do not count toward the threshold.** The alternative — excluding them from the window entirely — would make an old burst of timeouts linger indefinitely across a long run of connection errors. I think ageing them out is right, but it is a defensible either way.

Also worth a look: `isRecallTimeoutError` keeps a narrow anchored message fallback alongside the `instanceof` check, for timeouts that cross a boundary losing prototype identity. If you would rather drop the fallback entirely and rely on the type alone, say so.

## PR workflow

- [x] PR scope is narrow — one feature, confined to `packages/plugin-pi`
- [x] Synced with latest `main` before requesting AI review — branched from `main` at v9.28.0
- [x] Batched review fixes by subsystem instead of micro-pushing

🤖 *This content was generated with AI assistance using Claude Opus 5.*

## Summary by CodeRabbit

* **New Features**
  * Added an in-memory recall-timeout circuit breaker for Pi automatic between-tool recalls; once tripped, recall is disabled until restart.
  * Added `recallTimeoutThreshold` and `recallTimeoutWindow` configuration (documented and persisted), and exported the `PiApi` type.
* **Bug Fixes**
  * Improved caller abort handling: timeouts, retries/backoff, and in-flight recalls now cancel promptly and propagate abort errors correctly.
  * Recall tooling now consistently respects the breaker (skipping recall-producing tools and showing a “disabled until restart” notice).
* **Documentation**
  * Updated Pi plugin README and configuration examples for the new keys.
* **Tests**
  * Expanded validation, breaker rolling-window behavior, timeout classification, abort propagation, and recall-disable persistence across runs.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,18 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Added
+
+- Recall-timeout circuit breaker for the Pi plugin's automatic between-tool
+  context recalls (`packages/plugin-pi`). When `recallTimeoutThreshold` (default
+  `7`) of the last `recallTimeoutWindow` (default `10`) recall calls fail with an
+  explicit request timeout, automatic recall is disabled for the remainder of the
+  process: subsequent calls are neither retried nor awaited, in-flight recalls are
+  aborted, and the session status reports that recall is off until restart. The
+  breaker is in-memory, never re-enables itself after a cooldown, and adds no
+  latency on the healthy path. Only explicit request timeouts count toward
+  tripping; other failures still occupy window slots so stale timeouts age out.
+
 ## [v9.8.0] — 2026-07-21
 
 ### Added
--- a/packages/plugin-pi/README.md
+++ b/packages/plugin-pi/README.md
@@ -87,6 +87,8 @@ Supported config keys:
 | `statusEnabled` | `true` | Set Pi UI status from daemon health |
 | `requestTimeoutMs` | `60000` | HTTP/MCP request timeout for recall/observe/compaction/commands |
 | `startupRequestTimeoutMs` | `1000` | Shorter timeout for startup-sensitive probes (MCP `tools/list` registration and the `session_start` health/status check) so a slow or offline daemon can't stall Pi boot |
+| `recallTimeoutThreshold` | `7` | Permanently disable recall after this many explicit recall timeouts within the rolling window |
+| `recallTimeoutWindow` | `10` | Number of most recent recall calls included in the rolling timeout window |
 
 Boolean-like strings such as `"false"`, `"0"`, `"no"`, and `"off"` are treated as false.
 
@@ -107,7 +109,9 @@ Example:
   "mcpToolsEnabled": true,
   "statusEnabled": true,
   "requestTimeoutMs": 60000,
-  "startupRequestTimeoutMs": 1000
+  "startupRequestTimeoutMs": 1000,
+  "recallTimeoutThreshold": 7,
+  "recallTimeoutWindow": 10
 }
 ```
 
--- a/packages/plugin-pi/README.md
+++ b/packages/plugin-pi/README.md
@@ -87,8 +87,8 @@ Supported config keys:
 | `statusEnabled` | `true` | Set Pi UI status from daemon health |
 | `requestTimeoutMs` | `60000` | HTTP/MCP request timeout for recall/observe/compaction/commands |
 | `startupRequestTimeoutMs` | `1000` | Shorter timeout for startup-sensitive probes (MCP `tools/list` registration and the `session_start` health/status check) so a slow or offline daemon can't stall Pi boot |
-| `recallTimeoutThreshold` | `7` | Permanently disable recall after this many explicit recall timeouts within the rolling window |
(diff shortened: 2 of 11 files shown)


joshuaswarren:
Rebased onto current main and addressed the current review batch. Local Pi verification passes: `pnpm --filter @remnic/plugin-pi test` (166 passing) and `check-types`. The fresh CodeRabbit and Kilo reviews are green; GitHub Actions are awaiting maintainer approval for this fork PR. Please approve the pending workflows so the required gates can run on head `aa979bbb0`.

joshuaswarren:
Follow-up: addressed the credential-isolation review finding and pushed head `682e5d1c4`. The breaker now keys on a SHA-256 credential fingerprint, never the credential itself; regression coverage proves a separate credential gets a fresh breaker. Local Pi suite remains 166/166 passing. Pending GitHub Actions still require maintainer approval for this fork PR.

joshuaswarren:
@cursor review the current head for correctness and regressions.

joshuaswarren:
@cursor review the current head for correctness and regressions.

joshuaswarren:
@codex review current head 498143703 for correctness and regressions.

PR #2252, opened 2026-07-30
fix(core): make extraction diagnostics truthful

### LINKED ISSUE #2227: extraction liveness: exclude prefilter skips and defer the live stamp until downstream steps complete (#2223 follow-up)
## Summary

Follow-up to #2223 (PR #2224). The merged fix advances `lastExtractionAt` for any successfully parsed empty extraction. Two refinements came out of review that narrow "successful" so the liveness watermark reflects a genuine, fully-completed provider extraction rather than a skip or a pre-downstream stamp.

## 1. Prefilter skips must not count as a successful provider parse

`ExtractionEngine.extract()` returns empty arrays without invoking a provider in two short-circuit paths (`extraction.ts` ~L1162 and the mechanical-telemetry prefilter ~L1166-1177). Today those are indistinguishable from a real empty provider parse, so they advance the liveness watermark. Recurring telemetry can therefore keep `/health.extraction` fresh while genuine provider extractions are failing.

Proposal: mark short-circuit skips on the `ExtractionResult` (e.g. an optional `extractionSkippedReason`) and gate the liveness stamp on its absence.

## 2. For non-fingerprinted live runs, defer the diagnostic stamp until post-extraction steps complete

The empty-success liveness stamp runs before passive-correction capture and `clearAfterExtraction()`. If those fail, `runExtraction()` rejects and leaves the turns buffered, but the watermark is already refreshed. Repeated retries can keep the watermark fresh on a buffer that never drains, so `/health.extraction` reports a stuck pipeline as healthy.

Proposal: for the fail-open live path, stamp after the post-extraction steps complete successfully (the fingerprinted/import path keeps its current fatal, dedupe-load-bearing semantics).

## Scope

Both are diagnostics-accuracy refinements; neither changes the durable write path. Filed separately from #2223 so that fix stays narrowly scoped to the documented empty-provider-parse gap.

### LINKED ISSUE #2230: extraction liveness aggregate: namespace counts in stats, scoped-health display, timestamp validation (#2159 follow-ups)
## Summary

Follow-ups to #2159 (PR #2225), which shipped the aggregate extraction-liveness watermark across namespace stores. The watermark timestamp aggregation and partial-read-failure surfacing meet #2159's acceptance; these are adjacent edge cases the review surfaced that are out of scope for that change.

## 1. Namespace extraction counts absent from stats after default migration

When the default namespace has migrated off the legacy root, the aggregate watermark read skips `rootStorage` and only attaches `rootStats` from that single root read. Namespace `loadMeta` results never contribute counters, so `remnic stats` can show `Extractions: unavailable` even though the active namespace store has valid metadata. The doctor `rootMeta` is also ignored on the namespaced path. Aggregate namespace counters into `rootStats` (or a parallel summary).

## 2. Namespace-scoped health display consistency

For namespace-scoped capability tokens, `/health` computes liveness from the daemon-wide buffer but zeroes `bufferedSessionCount`/`pendingTurnCount`/`oldestBufferedTurnAgeMs` while leaving `degraded` unchanged, so a scoped caller can see `degraded: true` with an empty backlog. Decide one consistent scope for scoped tokens (capability-scoped buffer snapshot, or redact only the timestamp and keep the global counts) and apply it end to end.

## 3. Reject overflowed calendar timestamps

`Date.parse` normalizes some invalid calendar values (e.g. an impossible day) rather than rejecting them, so a corrupt `lastExtractionAt` could be accepted as a finite timestamp. Validate the timestamp round-trips (the daemon writes ISO via `toISOString()`, so this only affects corrupt/foreign stores).

## Scope

All three are diagnostics/stats accuracy and hardening refinements around the aggregate watermark; none changes the durable write path. Filed separately so #2159 stays scoped to the watermark-timestamp aggregation acceptance.

joshuaswarren:
## What changed

Local filter skips no longer count as provider runs.
Live health updates only after later work succeeds.
Stats now include every active namespace.
Scoped health keeps its backlog counts while hiding the exact run time.
Bad calendar dates now fail metadata checks.

## Checks

- `pnpm run check-types`
- `pnpm --filter @remnic/core test`
- `pnpm run build`
- `bash scripts/check-review-patterns.sh`
- `node scripts/check-ratchets.mjs`

Closes #2227
Closes #2230

---

> [!NOTE]
> **Medium Risk**
> Touches extraction orchestration, circuit-breaker healing, and health/watermark aggregation used for ops alerts; behavior changes could affect degradation signals and retry semantics, but changes are well covered by new tests.
> 
> **Overview**
> Makes **extraction health and stats** reflect what actually ran, not local prefilter skips or incomplete pipeline steps.
> 
> **Liveness and resilience:** Empty extractions now carry `extractionSkippedReason` (`conversation_only_non_memory`, `mechanical_telemetry`) when the provider is never called. `commitEmptySuccessExtraction` only bumps `extractionCount` / `lastExtractionAt` when there is no skip reason, and for live runs that stamp runs **after** passive capture and buffer clear so failures there do not look like success. Prefilter skips no longer reset the provider circuit breaker.
> 
> **Aggregate watermark / stats:** `readAggregateExtractionWatermark` merges `extractionCount` and newest `lastConsolidationAt` across root and namespace meta, drops partial-read survivor stats, and rejects invalid calendar ISO timestamps (including consolidation fields).
> 
> **Scoped `/health`:** Token-scoped callers still hide daemon `lastExtractionAt` but get clearer `degradedReason` text (watermark vs buffer vs backlog) instead of a generic pipeline message; tests also expect backlog session/turn metrics to remain visible for scoped health.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit db155121f0e5fccc9d7fdd241aa773d22e95c316. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

## Summary by CodeRabbit

- **New Features**
  - Extraction results that are intentionally skipped now include `extractionSkippedReason` (`conversation_only_non_memory` or `mechanical_telemetry`) for clearer diagnostics.

- **Bug Fixes**
  - Improved extraction health reporting with more specific scoped degradation reasons and corrected buffered/pending metrics.
  - Prevented liveness/watermark advancement for local skip and incomplete outcomes; refined timestamp validation and aggregate `rootStats` merging.

- **Tests**
  - Added/expanded deterministic timing, scoped health assertions, skipped-reason checks, stricter watermark/aggregation validation (including `rootStats`), and regression coverage for circuit-breaker behavior not being healed.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -6,7 +6,7 @@ All notable changes to this project will be documented in this file.
 
 ### Fixed
 
-- Extraction health no longer counts local filter skips as successful runs. Live empty runs now record health only after later steps finish.
+- Extraction health no longer treats local filter skips as successful runs. Live empty runs record health only when later steps succeed.
 - Stats now add extraction counts from active namespaces. Scoped health keeps the backlog counts behind its status. Invalid calendar dates now fail metadata checks.
 
 ## [v9.45.1] — 2026-07-30
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -4,6 +4,11 @@ All notable changes to this project will be documented in this file.
 
 ## [Unreleased]
 
+### Fixed
+
+- Extraction health no longer counts local filter skips as successful runs. Live empty runs now record health only after later steps finish.
+- Stats now add extraction counts from active namespaces. Scoped health keeps the backlog counts behind its status. Invalid calendar dates now fail metadata checks.
+
 ## [v9.45.1] — 2026-07-30
 
 ### Added
--- a/packages/remnic-core/src/access-service-health.test.ts
+++ b/packages/remnic-core/src/access-service-health.test.ts
@@ -474,10 +474,12 @@ test("health keeps namespace QMD failures scoped to the namespace", async () => 
   }
 });
 
-test("health surfaces extraction liveness as degraded when the buffer is non-empty and the watermark is stale", async () => {
+test("health surfaces extraction liveness as degraded when the buffer is non-empty and the watermark is stale", async (t) => {
+  const nowMs = Date.parse("2026-07-30T12:00:00.000Z");
+  t.mock.timers.enable({ apis: ["Date"], now: nowMs });
   const memoryDir = await mkdtemp(path.join(os.tmpdir(), "remnic-health-liveness-"));
   try {
-    const oldTs = new Date(Date.now() - 10_000).toISOString();
+    const oldTs = new Date(nowMs - 10_000).toISOString();
     const config = parseConfig({ memoryDir, qmdEnabled: false, extractionLiveness: { staleWindowMs: 1000 } });
     const service = new EngramAccessService({
       config,
@@ -506,11 +508,7 @@ test("health surfaces extraction liveness as degraded when the buffer is non-emp
     assert.equal(health.extraction.bufferedSessionCount, 3);
     assert.equal(health.extraction.pendingTurnCount, 12);
     assert.equal(health.extraction.lastExtractionAt, oldTs);
-    const oldestAge = health.extraction.oldestBufferedTurnAgeMs ?? -1;
-    assert.ok(
-      oldestAge >= 10_000 && oldestAge < 60_000,
-      `oldest buffered turn age reflects the ~10s offset, not merely non-negative (got ${oldestAge}ms)`,
-    );
+    assert.equal(health.extraction.oldestBufferedTurnAgeMs, 10_000);
     assert.ok(
       health.extraction.degradedReason !== null && health.extraction.degradedReason.length > 0,
       "degraded reason is populated",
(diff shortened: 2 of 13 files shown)


joshuaswarren (commented):
All current review findings have been fixed in code and validated locally. Please run the final review gate.

joshuaswarren (commented):
The breaker finding is fixed in code with a regression test, and all local gates now pass. Please run the final review gate.

joshuaswarren (commented):
Lowercase RFC 3339 markers are now supported with regression coverage. All local gates pass. Please run the final review gate.

joshuaswarren (commented):
Clean gate.

joshuaswarren (commented):
Clean gate after minute-precision fix.

joshuaswarren (commented):
Clean gate.

PR #2499, opened 2026-08-17
fix(extraction): salvage valid items from mixed-invalid model output

### LINKED ISSUE #2455: fix(extraction): salvage valid items from mixed-invalid model output
## Bug

`ExtractionResultSchema` validates each model-produced array as one strict unit. A single malformed fact, entity, question, relationship, or profile update rejects the entire extraction result and reduces otherwise valid output to `fallback_no_parsed_output`.

This is especially visible on gateway-routed models: one unsupported enum or one incomplete procedure fact discards every valid memory in the same response.

Observed failures included:

```text
facts.4.cueAnchors.0.type: invalid_enum_value
facts.3.procedureSteps: custom
facts.1.category: invalid_enum_value
entities.0.type: invalid_enum_value
entities.0.facts.0: invalid_type
```

## Impact

- valid memories in the same model response are lost;
- flush-plan replay repeatedly retries/preserves large tails;
- one accessory enrichment error can block durable-memory ingestion;
- bounded retries do not help deterministic schema deviations.

## Expected behavior

For mixed-validity arrays, retain elements that independently pass their item schema and discard only malformed elements. Preserve the current failure/retry behavior when a non-empty array is wholly invalid, so malformed output is not silently accepted as an empty extraction.

## Proposed behavior

- `valid + invalid` items -> parse succeeds with valid items;
- non-empty and all invalid -> parse fails, allowing the bounded retry;
- explicitly empty arrays -> remain valid;
- consolidation schemas remain strict and unchanged.

This keeps strict item contracts while moving failure isolation from the whole response to the malformed element.

rmichelena:
## Summary

- isolate schema failures to malformed items in mixed-validity extraction arrays
- retain valid facts, profile updates, entities, questions, and relationships from the same model response
- keep wholly-invalid non-empty arrays invalid so bounded retries still run instead of silently accepting an empty extraction
- preserve strict consolidation schemas

Fixes #2455

## Type of change

- [x] Bug fix
- [ ] Feature
- [ ] Docs
- [ ] Refactor
- [ ] Security / hardening

## Validation

- [x] Targeted extraction schema tests (2 passed)
- [x] Added tests for mixed-validity salvage and wholly-invalid retry behavior
- [x] `git diff --check`
- [ ] Full `npm test` (targeted schema coverage used for this narrow change)
- [ ] `npm run review:cursor` (Cursor CLI unavailable in this environment)

## Changelog

- [x] Changeset added for `@remnic/core` patch release

## Risk assessment

- [x] No secrets/tokens added
- [x] Backwards compatibility considered
- [x] Invalid user/model input remains rejected when no valid item can be recovered

## Notes for reviewers

The strict item schemas are unchanged. The preprocessors only change failure isolation:

- mixed valid/invalid arrays retain independently valid items;
- wholly-invalid non-empty arrays are passed through unchanged and fail normal Zod validation;
- empty arrays retain existing behavior.

This prevents one malformed enrichment from reducing an otherwise useful gateway extraction to `fallback_no_parsed_output` without weakening the retry path.

## PR workflow

- [x] PR scope is narrow
- [x] Synced with latest `main` before implementation
- [x] Batched review fixes by subsystem instead of micro-pushing

## Summary by CodeRabbit

- **Bug Fixes**
  - Preserved valid extraction results when gateway responses contain a mix of valid and invalid entries.
  - Discarded only invalid items across facts, profile updates, entities, questions, and relationships.
  - Continued retry handling for responses containing no valid items.
  - Applied consistent handling across standard and proactive extraction results.

### DIFF
--- a/.changeset/salvage-mixed-extraction-results.md
+++ b/.changeset/salvage-mixed-extraction-results.md
@@ -1,0 +1,5 @@
+---
+"@remnic/core": patch
+---
+
+Preserve valid extraction items when a gateway model returns mixed-validity result arrays while retaining retries for wholly invalid output.
--- a/packages/remnic-core/src/extraction-prompt-safety.test.ts
+++ b/packages/remnic-core/src/extraction-prompt-safety.test.ts
@@ -333,3 +333,59 @@ test("extraction schema permits empty optional question output", () => {
   assert.deepEqual(parsed.data?.questions, []);
   assert.match(ExtractionResultSchema.shape.questions.description ?? "", /zero to three/i);
 });
+
+test("extraction schema salvages valid items from mixed-validity model arrays", () => {
+  const parsed = ExtractionResultSchema.parse({
+    facts: [
+      {
+        category: "fact",
+        content: "Keep this valid memory.",
+        confidence: 0.95,
+        tags: ["salvage"],
+      },
+      {
+        category: "procedure",
+        content: "Malformed one-step procedure.",
+        confidence: 0.8,
+        tags: ["salvage"],
+        procedureSteps: [{ order: 1, intent: "Only one step" }],
+      },
+      {
+        category: "invented_category",
+        content: "Invalid enum value.",
+        confidence: 0.7,
+        tags: ["salvage"],
+      },
+    ],
+    profileUpdates: ["Keep this profile update.", { invalid: true }],
+    entities: [
+      { name: "qmd", type: "tool", facts: ["Valid entity fact."] },
+      { name: "invalid", type: "skill", facts: ["Invalid entity type."] },
+    ],
+    questions: [
+      { question: "Valid question?", context: "Schema fixture", priority: 0.5 },
+      { question: "Invalid priority", context: "Schema fixture", priority: 9 },
+    ],
+    relationships: [
+      { source: "remnic", target: "qmd", label: "uses" },
+      { source: "missing-fields" },
+    ],
+  });
+
+  assert.deepEqual(parsed.facts.map((fact) => fact.content), ["Keep this valid memory."]);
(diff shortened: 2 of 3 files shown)


joshuaswarren:
Thanks for this.

#2455 already shipped on `main` as `salvageArray()` in `packages/remnic-core/src/schemas.ts`. Mixed-validity extraction arrays already keep valid items and reject a wholly-invalid non-empty array so retries still run.

This branch is behind current `main`. A two-dot diff would replace that helper and delete later landings, not add a remaining salvage path.

joshuaswarren (changes requested):

joshuaswarren:
Adversarial follow-up on the unique hunk, not only the stale-base problem:

`salvageArrayItems` / `salvageStringItems` return `[]` when the field is missing, null, or not an array. The schema then accepts that empty list. A model object that omits `facts` or sends a non-array can parse as a successful empty extraction, so bounded retry never runs.

Current `main` `salvageArray()` still requires a real array. Do not replace it. If you want salvage on `ProactiveExtractionResultSchema`, open a new PR against current `main` that reuses `salvageArray` and leaves `ExtractionResultSchema` alone.

rmichelena:
Thanks, Joshua — confirmed, including the adversarial case you highlighted. Current `main`'s `salvageArray()` preserves the required-array validation boundary while salvaging valid members from mixed arrays. This branch's preprocessors would incorrectly turn missing or non-array required fields into `[]`, potentially bypassing bounded retry, so there is no useful remaining delta here.

Apologies for opening this PR after #2464 had landed. We prepared these changes a couple of days ago, but an authentication-token problem delayed publishing them and we missed how quickly `main` had advanced. Congratulations on landing a more robust solution so quickly.

Closing this as superseded by #2464.

PR #2500, opened 2026-08-17
fix(qmd): kill subprocess descendants on timeout and abort

### LINKED ISSUE #2456: fix(qmd): kill subprocess descendants on timeout and abort
## Bug

QMD subprocess timeout/abort cleanup calls `child.kill("SIGKILL")`, which terminates only the launcher process. QMD can spawn model, embedding, or runtime descendants; those grandchildren survive the timeout and retain memory/CPU after the caller has already failed.

Current paths in `packages/remnic-core/src/qmd.ts`:

```ts
const child = launchProcess(command, args, { ... });
// timeout
child.kill("SIGKILL");
// abort
child.kill("SIGKILL");
```

## Observed impact

After repeated QMD latency timeouts, orphaned embedding/model grandchildren accumulated and memory usage continued growing even though the owning QMD command had failed. Killing the process group eliminated the descendants in a controlled reproduction.

## Expected behavior

On POSIX systems, each bounded QMD subprocess should run in its own process group. Timeout/abort cleanup should signal the process group, with child-only fallback for unsupported platforms and exit races.

## Proposed fix

- launch bounded QMD commands with `detached: true` on POSIX;
- on timeout/abort, call `process.kill(-child.pid, signal)`;
- retain `child.kill(signal)` as the Windows/failure fallback;
- add an integration regression test where a fake QMD launcher spawns a grandchild and times out, asserting the grandchild is no longer running.

The long-lived MCP daemon lifecycle is unchanged; this fix applies only to bounded subprocess commands.

rmichelena:
## Summary

- launch bounded QMD subprocesses in their own POSIX process group
- terminate the full process group on timeout or abort, with child-only fallback
- prevent model and embedding descendants from remaining orphaned
- leave the long-lived QMD MCP daemon lifecycle unchanged

Fixes #2456

## Validation

- [x] Integration regression test: fake QMD launcher spawns a grandchild; timeout removes both
- [x] `git diff --check`
- [x] Changeset added for `@remnic/core`

## Risk assessment

The group signal is used only for QMD commands launched with `detached: true` on POSIX. Windows and group-signal failures retain the existing direct-child fallback.

### DIFF
--- a/.changeset/kill-qmd-process-groups.md
+++ b/.changeset/kill-qmd-process-groups.md
@@ -1,0 +1,5 @@
+---
+"@remnic/core": patch
+---
+
+Terminate QMD subprocess process groups on timeout or abort so model and embedding descendants cannot remain orphaned.
--- a/packages/remnic-core/src/qmd.test.ts
+++ b/packages/remnic-core/src/qmd.test.ts
@@ -1,4 +1,5 @@
 import assert from "node:assert/strict";
+import { chmod, mkdtemp, readFile, rm, writeFile } from "node:fs/promises";
 import os from "node:os";
 import path from "node:path";
 import test from "node:test";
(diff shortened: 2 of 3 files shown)


joshuaswarren:
Thanks for this.

#2456 already shipped on `main`. `qmd.ts` already launches POSIX subprocesses `detached` and kills the process group with `process.kill(-pid)` on timeout/abort.

This branch is behind current `main`. A two-dot merge would delete the existing process-group test module and other later landings.

joshuaswarren (changes requested):

rmichelena:
Thanks, Joshua — confirmed. Current `main` already covers both timeout and abort, using a detached POSIX process group, negative-PID signaling, and a direct-kill fallback, with regression tests that verify grandchildren are terminated.

Apologies for opening this PR after #2460 had landed. We prepared these changes a couple of days ago, but an authentication-token problem delayed publishing them and we missed the intervening progress on `main`. Congratulations on moving the fix through so quickly.

Closing this as superseded by #2460.

PR #2501, opened 2026-08-17
fix(openclaw): bypass LCM for flush-plan recovery imports

### LINKED ISSUE #2457: fix(openclaw): bypass LCM for flush-plan recovery imports
## Bug

OpenClaw flush-plan recovery turns are already condensed compaction output, but `ingestBulkImportBatch` sends them through `lcmEngine.observeMessages` before extraction. This unnecessarily feeds recovery material back into LCM, adds latency and memory pressure, and can recursively summarize content that is itself the result of compaction.

Flush-plan turns already carry trusted provenance:

- `sourceLabel: "OpenClaw flush plan"`
- `sourceId: "<service>:flush-plan[:chunk]"`

Despite that, `packages/remnic-core/src/orchestration/turn-ingestion.ts` currently observes every bulk import unconditionally.

## Expected behavior

Skip LCM observation only for batches identified by the existing OpenClaw flush-plan provenance. Continue normal LCM observation for all ordinary bulk imports. Extraction and persistence behavior must remain unchanged.

## Proposed fix

- detect flush-plan provenance before `lcmEngine.observeMessages`;
- bypass observation for those recovery batches;
- add regression tests proving flush-plan imports skip LCM and ordinary imports still observe.

rmichelena:
## Summary

- identify OpenClaw flush-plan recovery batches using their existing trusted provenance
- skip LCM observation for those already-compacted recovery turns
- preserve normal LCM observation for every ordinary bulk import
- leave extraction and persistence behavior unchanged

Fixes #2457

## Validation

- [x] Flush-plan import skips LCM observation
- [x] Ordinary import still invokes LCM observation
- [x] `git diff --check`
- [x] Changeset added for `@remnic/core`

## Risk assessment

The bypass is narrowly gated by `sourceLabel === "OpenClaw flush plan"` or the established `:flush-plan` source identifier. Other imports retain existing behavior.

### DIFF
--- a/.changeset/skip-flush-plan-lcm.md
+++ b/.changeset/skip-flush-plan-lcm.md
@@ -1,0 +1,5 @@
+---
+"@remnic/core": patch
+---
+
+Skip LCM observation for OpenClaw flush-plan recovery imports while preserving observation for ordinary bulk imports.
--- a/packages/remnic-core/src/orchestration/turn-ingestion.ts
+++ b/packages/remnic-core/src/orchestration/turn-ingestion.ts
@@ -456,7 +456,12 @@ export class TurnIngestionCoordinator {
       };
     }
 
-    if (this.deps.lcmEngine?.enabled) {
+    const isOpenClawFlushPlanImport = sessionTurns.some((turn) => {
+      const provenance = turn.importProvenance;
+      return provenance?.sourceLabel === "OpenClaw flush plan" ||
+        provenance?.sourceId?.includes(":flush-plan") === true;
+    });
+    if (this.deps.lcmEngine?.enabled && !isOpenClawFlushPlanImport) {
       await this.deps.lcmEngine.observeMessages(
         sessionKey,
         sessionTurns.map((turn) => ({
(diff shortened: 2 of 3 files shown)


joshuaswarren:
Thanks for this.

#2457 already shipped on `main` with a stricter gate: skip LCM only when every turn is flush-plan recovery (`sessionTurns.every(isFlushPlanRecoveryTurn)` plus a `:flush-plan(:n/n)?` sourceId check).

The change here uses `.some()` on a looser sourceId `includes` check. That is weaker than `main` (a mixed batch would skip LCM for ordinary turns). It is not an additional fix.

This branch is also behind current `main`. Please close this PR, or rebase onto current `main` and keep only a delta that is not already shipped. I will not merge it as-is.

joshuaswarren (changes requested):

rmichelena:
Thanks, Joshua — confirmed. Current `main` is stricter and safer: it skips LCM only when every turn is flush-plan recovery material, and its source-id match is anchored to the expected flush-plan suffix. This branch's `some()` plus loose `includes()` check could incorrectly skip ordinary turns in a mixed batch, so it should not be adapted or rebased.

Apologies for opening this PR after #2463 had landed. We prepared these changes a couple of days ago, but an authentication-token problem delayed publishing them and we missed how quickly `main` had advanced. Congratulations on landing the stronger implementation so quickly.

Closing this as superseded by #2463.

PR #2503, opened 2026-08-17
fix(openclaw): replay flush-plan chunks in isolated serial batches

### LINKED ISSUE #2459: fix(openclaw): replay flush-plan chunks in isolated serial batches
## Bug

Flush-plan recovery chunks an oversized snapshot into multiple `ImportTurn`s, but then submits all turns in one `ingestBulkImportBatch` call. Because the turns carry persistent fingerprints, bulk ingestion gives the entire batch one stable session key and shared extraction buffer. This recombines the chunks that were split to bound extraction size.

## Observed impact

Large compaction recovery plans become oversized extraction prompts, increasing model timeouts, heap pressure, and partial-replay latency. A failure late in the combined batch also delays durable progress for earlier chunks.

## Expected behavior

Replay flush-plan turns serially, one turn per bulk-ingest call. Each chunk then receives an isolated import session while preserving order, aggregate counters, existing deadline semantics, and processed-prefix/tail recovery.

## Proposed fix

- call `ingestBulkImportBatch([turn])` sequentially;
- aggregate ingest counters across successful turns;
- stop at the first failed result;
- when a later call throws, return an aggregate partial result so the already-processed prefix is removed and only the tail is retained;
- add a regression assertion proving every oversized-plan ingest batch has size one.

rmichelena:
## Summary

- replay flush-plan chunks serially, one turn per bulk-import batch
- prevent chunked recovery material from being recombined into one shared extraction session
- aggregate ingest counters across successful chunks
- stop on the first failed chunk and preserve only the unprocessed tail

Fixes #2459

## Validation

- [x] Oversized-plan regression test verifies every ingest batch contains exactly one turn
- [x] Existing chunking and cleanup behavior remains green
- [x] `node scripts/pnpm.mjs --filter @remnic/plugin-openclaw run check-types`
- [x] `git diff --check`
- [x] Changeset added for `@remnic/plugin-openclaw`

## Risk assessment

Ordering remains serial and existing fingerprint, marker, partial-prefix, deadline, and tail-preservation paths are reused. The change reduces extraction batch size rather than introducing concurrency.

### DIFF
--- a/.changeset/serialize-flush-plan-replay.md
+++ b/.changeset/serialize-flush-plan-replay.md
@@ -1,0 +1,5 @@
+---
+"@remnic/plugin-openclaw": patch
+---
+
+Replay OpenClaw flush-plan chunks serially so each extraction uses an isolated import session.
--- a/src/openclaw-flush-plan-lifecycle.ts
+++ b/src/openclaw-flush-plan-lifecycle.ts
@@ -789,25 +789,51 @@ async function ingestFlushPlanImportTurns(params: {
   importTurns: ImportTurn[];
   deadlineMs?: number;
 }): Promise<void | OpenClawFlushPlanIngestResult> {
-  try {
-    return await params.ingestor.ingestBulkImportBatch(params.importTurns, {
-      ...(params.deadlineMs === undefined
-        ? {}
-        : { deadlineMs: params.deadlineMs }),
-      failOnExtractionFailure: true,
-      includeSourceValidAtContext: false,
-    });
-  } catch (error) {
-    const partialResult = partialIngestResultFromError(error);
-    if (
-      partialResult &&
-      typeof partialResult.processedTurnCount === "number" &&
-      partialResult.processedTurnCount > 0
-    ) {
-      return partialResult;
+  const aggregate: Required<OpenClawFlushPlanIngestResult> = {
+    attemptedTurnCount: 0,
+    extractionCount: 0,
+    persistedCount: 0,
+    durableOutputCount: 0,
+    skippedCount: 0,
+    failedCount: 0,
+    postPersistMetadataFailureCount: 0,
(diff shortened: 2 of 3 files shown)


joshuaswarren:
Thanks for this.

#2459 already shipped on `main` as `src/openclaw-flush-plan-serial-ingest.ts`. Replay already ingest one turn per `ingestBulkImportBatch` call, aggregates counters, and stops on the first failed chunk.

This branch is behind current `main`. A two-dot merge would delete that helper and re-inline a less complete copy.

joshuaswarren (changes requested):

rmichelena:
Thanks, Joshua — confirmed, including the two review findings. Current `main`'s extracted serial-ingest helper uses a typed counter-key list and correctly handles an explicit `processedTurnCount: 0`, while preserving the processed prefix and retaining only the failed tail. This branch is both stale and less complete, so there is no useful delta to rebase.

Apologies for opening this PR after #2461 had landed. We prepared these changes a couple of days ago, but an authentication-token problem delayed publishing them and we missed how quickly `main` had advanced. Congratulations on landing the more complete implementation so quickly.

Closing this as superseded by #2461.

joshuaswarren:
@rmichelena - no worries, and sorry you put in the work on something that got merged by one of my own agents earlier. I was excited to see so many PRs come in from you! 

Let me know if you're planning on doing more work on Remnic, I'd be more than happy to setup a way that you can see what my agents currently have in flight so that you don't duplicate it, or mark issues that I know I'm not going to be getting to for a bit. 

Overall, though, I've got a new Grok SuperHeavy subscription added to the mix, so I'm doing a major push to burn down the number of open issues on Remnic over the next few days, so there may be a lot of churn on the issues. Maybe there's a new feature you could work on that would help you with your use case and then PR?

PR #2545, opened 2026-08-18
fix(extraction): request native structured output

### LINKED ISSUE #2544: Gateway extraction lacks a structured JSON output contract
## Summary

Gateway-backed extraction does not include the extraction response shape in its prompt and does not request provider-native structured output. As a result, otherwise capable models can return plausible but incompatible object/null/string shapes for `facts`, `profileUpdates`, or `entities`, and `ExtractionResultSchema` rejects the response.

The direct and local extraction paths include `EXTRACTION_RESPONSE_SHAPE`; the gateway fallback path currently sends only `buildExtractionInstructions(...)`.

## Impact

- Recall remains healthy, but background extraction can produce zero memories.
- Schema-validation retries cannot recover reliably because the retry refers to a shape that was never included in the gateway prompt.
- Slow task-model chains can exhaust the extraction deadline while retrying malformed output.

Observed validation failures include:

```text
facts:invalid_type, profileUpdates:invalid_type, entities:invalid_type
```

## Expected behavior

The gateway path should:

1. Send native JSON Schema constraints when the selected OpenAI-compatible transport supports them.
2. Translate the same option to `response_format.json_schema` for Chat Completions and `text.format.json_schema` for Responses API.
3. Retry once without native schema only when the provider explicitly rejects structured output.
4. Keep the textual response shape as a compatibility fallback.
5. Retain the existing tolerant Zod parser as the final validation/salvage boundary.

## Proposed implementation

Use a transform-free provider-facing extraction schema for JSON Schema generation. The current tolerant extraction schema contains salvage effects that cannot be represented faithfully and can otherwise collapse array item schemas.

A patch with transport tests and extraction wiring is ready.

rmichelena:
## Summary

- add a transform-free provider-facing extraction schema and generate JSON Schema from it
- request native structured output on both Chat Completions and Responses API transports
- retry the same model once without native schema when an OpenAI-compatible provider rejects structured output
- include the existing extraction response shape in the gateway prompt as a compatibility fallback
- preserve tolerant `ExtractionResultSchema` validation as the final boundary

Fixes #2544.

## Why

The gateway extraction path only passed `buildExtractionInstructions(...)`; unlike the local/direct paths, it omitted `EXTRACTION_RESPONSE_SHAPE`. Models could therefore return plausible but incompatible shapes, causing `facts:invalid_type`, `profileUpdates:invalid_type`, and `entities:invalid_type` validation failures.

The provider-facing schema is deliberately separate from `ExtractionResultSchema`: the latter contains salvage transforms that JSON Schema cannot represent faithfully.

## Transport behavior

- Chat Completions: `response_format: { type: "json_schema", json_schema: ... }`
- Responses API: `text.format: { type: "json_schema", ... }`
- Unsupported 400/404/422 response: retry once without native schema, then continue through the existing model chain
- Non-schema provider errors: no special retry

## Validation

- `pnpm --filter @remnic/core check-types`
- `pnpm --filter @remnic/core build`
- focused extraction/fallback tests: 38 passed
- full core suite: 8308 passed, 32 skipped; one pre-build compatibility test failed because `dist/cli-fallback.js` did not yet exist
- after building, the compatibility test passed: 5/5
- generated extraction schema inspected: typed object/array items, no empty `items: {}` and no OpenAPI-only `nullable`

## Summary by CodeRabbit

* **New Features**
  * Improved extraction reliability with structured JSON responses from supported AI providers.
  * Added consistent validation for extracted facts, profile updates, entities, questions, and relationship data.
  * Added native JSON Schema support across compatible provider request modes.

* **Bug Fixes**
  * Providers that reject structured-output requests now retry without schema formatting and continue fallback processing.
  * Improved consistency between extraction instructions and returned data formats.


rmichelena:
**PR:** joshuaswarren/remnic#2545
**Head:** `b0fee08a5` (branch `fix/native-structured-extraction`)
**Base:** `main` (`721dd28e9`)
**Diff:** 7 files, +168/−8 (337 lines)
**Reviewers:** GPT-5.6 Terra, GLM 5.3, Qwen 3.7 Max

---

## Findings

### 🔴 High

*None found.* The PR's core architecture (dual schema, schema-less degradation, acceptResponse semantics) is correct.

### 🟠 Medium (5)

#### M1 (3/3 convergent) — `isUnsupportedJsonSchemaError` regex over-matches on bare "unsupported"

**File:** `packages/remnic-core/src/fallback-llm.ts:113`

The regex `/(?:400|404|422).*?(?:response[_ ]?format|json[_ ]?schema|structured output|unsupported)/i` treats "unsupported" as a standalone alternative. Any 400/404/422 mentioning "unsupported" for an unrelated reason (unsupported parameter, unsupported model, unsupported region) triggers the schema-stripping retry — doubling latency/cost and mis-diagnosing the root cause.

**Fix:** Drop the bare "unsupported" alternative, or require adjacency to schema context. Parse HTTP status structurally rather than from the error message string.

#### M2 (GLM) — Per-model reject retry re-paid on every chain walk, no capability memo

**File:** `packages/remnic-core/src/fallback-llm.ts:242`

When a provider rejects response_format/text.format, jsonSchema is never cleared for subsequent models. N-model chain where ALL reject structured output → 2N calls per extraction, compounded by bounded-retry loops.

**Fix:** Clear jsonSchema for remainder of chain after first detection. Consider per-provider capability cache.

#### M3 (GLM) — `options.jsonSchema` silent no-op on anthropic/ollama/codex transports

**File:** `packages/remnic-core/src/fallback-llm.ts:771`

Only wired into callOpenAI and callOpenAIResponses. Silent degradation on other transports. Ollama's /api/chat supports a `format` field.

**Fix:** Map to Ollama's format field, log debug note on Anthropic/Codex, document transport matrix.

#### M4 (Qwen) — Retry-path acceptResponse rejection logs wrong error

**File:** `packages/remnic-core/src/fallback-llm.ts:248-262`

When schema-rejection retry succeeds at transport level but acceptResponse() rejects, the catch block logs the *original* schema-rejection error. Forensics point to wrong cause.

**Fix:** Log distinct message: "fallback LLM: unstructured retry output rejected by acceptResponse"

#### M5 (Qwen) — `reasoningTrace.refine()` lost in JSON Schema — silent fact drops

**File:** `packages/remnic-core/src/schemas.ts:163-168`

effectStrategy:'input' strips .refine() that requires finalAnswer/final_answer. Provider can emit neither field → passes provider schema but fails ExtractionResultSchema → salvageArray drops fact silently.

**Fix:** Add description to reasoningTrace field stating "Must include finalAnswer or final_answer". Consider post-processing to inject oneOf constraint.

### 🟢 Low (9)

- **L1 (GLM)** — jsonSchema overrides responsesJsonSchema when both set, flipping strict
- **L2 (GLM)** — Module-load-time zodToJsonSchema can crash plugin import
- **L3 (GLM)** — zod specifier ^3.24.0 below zod-to-json-schema peer floor ^3.25.28
- **L4 (GLM)** — Three extraction shapes with no drift guard
- **L5 (GLM)** — No test coverage for Responses-API wiring or acceptResponse-rejected path
- **L6 (Qwen)** — Responses API strict default inconsistency
- **L7 (Qwen)** — EXTRACTION_RESPONSE_SHAPE always appended on happy path
- **L8 (Qwen)** — Double-billing risk on schema-rejection retry
- **L9 (Qwen)** — zod-to-json-schema pinned without caret

---

## Verdict

**Mergeable with follow-ups.** Clean dual-schema architecture, correct effectStrategy handling, correct strict:false pinning, working retry degradation. No critical or high-severity issues.

**Top fixes before production deployment:** M1 (3/3 convergent regex), M2 (chain-wide retry amplification), M3 (transport matrix), M5 (refine gap), L4 (drift guard).

rmichelena:
I have to say, this fixed a long-standing pain with the way I use Remnic (via the gateway's LLM providers, specifically DeepSeek 4 Flash) - it's made quite a difference.

Issue #2050, opened 2026-07-19
Dayflow activity-card AI analysis with strict evidence-bound output


joshuaswarren:
**Depends on:** #2048 and #2049
**Phase:** 2

## Goal

Add provider-neutral, optional AI analysis that turns deterministic activity observations/cards into concise, evidence-linked timeline narratives. Preserve deterministic cards when AI is unavailable or wrong.

## Scope

Implement an activity-analysis service in `packages/remnic-core` using existing model/provider registries, secret references, timeout/abort helpers, JSON extraction/validation, and logging rules.

The contract must accept a bounded day/window containing:

- timezone and date;
- observations and existing deterministic cards;
- category definitions;
- user preferences and prior accepted card edits.

The prompt must instruct the provider to use only supplied evidence, preserve chronology, avoid invented people/places/tasks, avoid productivity or emotional claims, merge adjacent activity only when justified, and return strict JSON. The output must contain validated card operations with evidence ranges, title, summary, category, confidence, and uncertainty. Include provider/model/prompt-version metadata without logging prompts or response content.

Batch adjacent observations with a bounded size and explicit overlap rules. Validate the complete result before applying it. A malformed, partial, aborted, timed-out, or provider-failed response must leave the deterministic timeline intact and record a typed failure status. Do not silently fall back from an explicitly invalid provider/model/config value.

Support local providers and configured remote providers through existing contracts. AI analysis is separately gated from activity capture and memory creation. No raw screenshots, audio, OCR output, keystrokes, clipboard contents, or media files may be sent by this feature.

## Non-goals

- Capture daemon or activity store changes.
- Category first-pass classification.
- Daily journal/weekly analytics generation.
- Automatic durable memory creation.

## Acceptance criteria

- [ ] Provider contract tests cover local/remote selection, missing provider, invalid config, timeout, abort, rate limit, malformed JSON, invalid schema, partial output, and empty input.
- [ ] Prompt snapshots prove evidence-only, chronology, no-invention, and strict-output instructions are present without user secrets/content in logs.
- [ ] Batch limits and deterministic ordering are enforced.
- [ ] Valid output updates only allowed cards/fields and preserves manual edits/source evidence.
- [ ] Any failure leaves deterministic cards unchanged and exposes a typed status.
- [ ] AI-disabled configuration makes zero provider calls and zero analysis artifacts.
- [ ] Tests use synthetic observations and pass `npm run preflight:quick`.

PR #2716, opened 2026-08-19
feat(activity): add isEmptyAnalysisBatches helper

joshuaswarren:
Part of #2050.


joshuaswarren:
Fixed in 9b1a515dd: inner null/undefined batches treated as empty; added regression tests for [null], [undefined], and [null, []].

Issue #2333, opened 2026-08-08
[harmonic 5/5] Span-mode extraction: offset-based memory values (bench-gated, two-phase)


joshuaswarren:
Part of the **harmonic-memory** series (labels: `harmonic-memory` + `priority:P1`..`priority:P5`). This issue is **P5 — implement LAST, and it is bench-gated**: Phase B (production) only proceeds if Phase A (benchmark) clears the numeric bar below. Independent of the other four issues.

### Series

| Order | Issue | Depends on |
|---|---|---|
| P1 | #2329 Construction pipeline | — |
| P2 | #2330 Merge-on-write | — (independent, but shares `extraction-persist.ts` touchpoints with #2329 — rebase carefully if both are in flight) |
| P3 | #2331 Episode-grouped recall | — |
| P4 | #2332 Deep-recall surface | **#2329 must merge first** (needs a populated anchor graph) |
| P5 | #2333 Span-mode extraction | — (bench-gated; do last) |

Order rationale: P1 makes the already-shipped harmonic retrieval tier functional and unblocks P4. P2 is the paper's largest measured quality gain. P3 is cheap and independent. P4 needs P1's data to be worth anything. P5 changes the extraction contract and only proceeds on benchmark evidence.

## Background

The Memora paper ([arXiv 2602.03315](https://arxiv.org/abs/2602.03315), §5.2.4, Table 6) reports a construction-cost optimization: instead of having the extraction LLM **generate** each memory value as free text, it predicts **index offsets into the source material**, and the value is materialized by slicing the source. This cut memory-construction time 45% (1322s → 740s per LoCoMo conversation) with near-zero quality loss (0.863 → 0.860). Generation tokens dominate extraction cost, and offsets are far fewer output tokens than full restatements.

Why this matters for Remnic: extraction is the dominant LLM cost of the plugin, and deployments on small local models (the `fallback-llm.ts`/`local-llm.ts` path) pay it in wall-clock. Halving output tokens also reduces hallucinated paraphrase drift — the value is verbatim source text.

Why this is P5 and bench-gated: it changes the extraction contract, and Remnic's quality safeguards key off generated content — the content-hash dedup index (Checklist #13), the semantic dedup lookup, the extraction judge, and grounding checks all consume `fact.content`. A naive port could regress dedup or recall quality invisibly. So: measure first, productionize second.

[...]

PR #2759, opened 2026-08-20
feat(extraction): stamp and verify the span source string to catch offset drift

joshuaswarren:
## Summary

Implements #2333's corruption guard: "Offsets must index the exact normalized string sent to the model — hash that string and assert the hash matches at materialization time (offset drift = silent corruption; reject on mismatch)."

This is the check that makes span mode safe to trust. If the text is re-normalized between prompt and materialization, the offsets still resolve — they just point at different characters, and a wrong quote gets persisted as verbatim source with no error anywhere.

`stampSpanSource` takes a SHA-256 over the exact bytes, with no trimming or newline normalization, because byte-identity is the entire property being asserted. `verifySpanSource` reports `length_mismatch` before comparing digests so the failure names the cheaper, more diagnostic dimension first, and a same-length single-character edit still surfaces as `hash_mismatch`.

Neither the result nor any error message carries the text: the source string is user content, and this stamp exists precisely so that content does not have to travel. A malformed `expected` stamp throws rather than comparing unequal, so a corrupt stamp cannot masquerade as drift.

Part of #2333 (Phase A guard; no Phase B wiring, `extraction.ts` untouched, so the issue stays open).

## Test plan

- `NODE_OPTIONS=--conditions=remnic-source npx tsx --test packages/remnic-core/src/extraction-span-source-hash.test.ts` — 11/11
- **Mutation-checked**: disabling the digest comparison (length-only verification) fails the same-length-edit test, so drift detection is genuinely asserted.
- Covers the empty string, both mismatch kinds, malformed stamps (wrong length, uppercase, non-hex, non-string, negative length), and that a failure result leaks neither the original nor the drifted text.

## Summary by CodeRabbit

* **New Features**
  * Added source stamping for extracted spans using a secure hash and recorded text length.
  * Added verification that detects source length or content mismatches with structured results.
  * Added validation for invalid or malformed source stamps without exposing source text.

* **Bug Fixes**
  * Improved handling of empty text and malformed verification data.

### DIFF
--- a/.changeset/2333-span-source-hash.md
+++ b/.changeset/2333-span-source-hash.md
@@ -1,0 +1,5 @@
+---
+"@remnic/core": patch
+---
+
+Add a span source stamp and verify helper. The stamp records the sha256 and length of the exact string sent to the model; verification compares length first, then the digest, and reports `length_mismatch` or `hash_mismatch` with both stamps. Malformed stamps throw before comparison. The helper is exported from `@remnic/core`; extraction wiring is unchanged. Part of #2333.
--- a/.changeset/2333-span-source-hash.md
+++ b/.changeset/2333-span-source-hash.md
@@ -2,4 +2,4 @@
 "@remnic/core": patch
 ---
 
-Add a span source stamp and verify helper. The stamp records the sha256 and length of the exact string sent to the model; verification compares length first, then the digest, and reports `length_mismatch` or `hash_mismatch` with both stamps. Malformed stamps throw before comparison. The helper is exported from `@remnic/core`; extraction wiring is unchanged. Part of #2333.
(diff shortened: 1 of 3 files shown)


joshuaswarren:
**Round 1 fixes landed in 9cc8a60a3.** All three findings were real; the first two defeated the guard this module exists to be.

- **Unpaired surrogates.** Confirmed empirically before fixing: under `utf8`, `"\ud800"`, `"\ud801"`, and `"\ufffd"` all hash to the same digest, so drift between them verified as unchanged. The digest now covers UTF-16LE code units, which is lossless for any JavaScript string. Regression asserts all three stamps differ and that the drift case is caught. No stamps are persisted yet, so the algorithm change costs nothing downstream.
- **Stamp projection.** `verifySpanSource` echoed the caller's object, so an extra `source` property would have travelled with the failure — the exact opposite of the no-content contract. The result now projects to exactly `hash` and `length`; the regression puts a sensitive-looking property on the expected stamp and asserts the serialized result contains neither it nor the key.
- **Export claim.** The changeset implied a package-level export. This module is source-internal like its fifteen `extraction-span-*` siblings, and the changeset now says so rather than adding an export the family does not have.

Both new tests fail on the pre-fix tree.

Issue #2052, opened 2026-07-19
Dayflow weekly activity analytics and dashboard snapshot


joshuaswarren:
**Depends on:** #2048 and #2049
**Phase:** 4

## Goal

Add deterministic weekly analytics over timeline cards. Produce a machine-readable snapshot that CLI/MCP/HTTP and a future UI can consume without making unsupported productivity claims.

## Scope

Implement a weekly analytics model and builder in `packages/remnic-core` with:

- configured IANA timezone, week start, and explicit half-open week range;
- totals by every configured category, including zero totals;
- active, idle, pause, gap, and unclassified durations;
- day-by-day totals and previous-period comparison;
- top application/domain summaries derived from structured activity metadata;
- recurring category/application patterns only when a deterministic threshold is met;
- source card/evidence counts and uncertainty flags;
- stable category ordering and deterministic tie-breakers.

Use timeline cards from #2049 as the only input. Do not infer mood, productivity, intent, presence, or quality of work. Call the result a time/activity summary, not a score.

Persist a versioned snapshot keyed by namespace, week range, source revision, and configuration hash. Rebuilding the same inputs must be idempotent. A changed category definition or source revision must produce a new snapshot rather than mutating an unrelated period. Retention/deletion must remove snapshots under the activity policy.

## Non-goals

- LLM-generated analysis or prose; that belongs to the journal layer.
- Capture/storage changes, screenshot/audio access, or dashboards/UI.
- Ranking people or evaluating productivity.

## Acceptance criteria

- [ ] Synthetic fixtures cover empty weeks, partial weeks, exact boundaries, DST, overlaps, gaps, idle/pause, unknown categories, deleted categories, and zero-valued categories.
- [ ] Week/day ranges are `[start,end)` and never double-count a boundary event.
- [ ] Category/application output ordering is stable for equal totals.
- [ ] Previous-period comparison is explicit when unavailable, not silently treated as zero.
- [ ] Snapshots are namespace-scoped, versioned, idempotent, and source-revision-aware.
- [ ] No unsupported productivity/mood/intent claims appear in output or docs.
- [ ] Tests pass with `npm run preflight:quick`.

PR #2760, opened 2026-08-20
feat(activity): compute an order-stable weekly snapshot config hash

joshuaswarren:
## Summary

Computes the `configHash` that #2052's snapshot key already requires. `weekly-persist.ts` has consumed a `configHash` since it landed, but nothing produced one — so every caller would have invented its own, and two callers would have disagreed, silently splitting or colliding stored snapshots.

`computeWeeklyConfigHash` is stable across category array order and object key insertion order, because a harmless reorder must not orphan every stored snapshot. It changes for every semantic change the issue names: renamed category, added category, removed category, different timezone, different week start — each tested individually.

Duplicate category ids throw rather than hashing, since two definitions for one id make the snapshot key ambiguous. The digest length is asserted in a test so a future algorithm swap is a visible, deliberate change rather than a silent one that invalidates stored snapshots.

Part of #2052 (hash helper; the persist call site passes its own value today, so wiring is a later slice and the issue stays open).

## Test plan

- `NODE_OPTIONS=--conditions=remnic-source npx tsx --test packages/remnic-core/src/activity/timeline/week-config-hash.test.ts` — 12/12
- Covers order-independence on both axes, all five semantic changes, duplicate ids, blank field rejection, digest shape, and that the caller's array is not sorted in place.

## Summary by CodeRabbit

* **New Features**
  * Added an exported helper for generating stable hashes from weekly configuration settings, including timezone, week start, and categories.
  * Hashes remain consistent regardless of category ordering and use a standardized lowercase format.
  * Added validation for invalid settings, duplicate category IDs, and blank fields.
* **Bug Fixes**
  * Preserves the original category list while processing configuration data.

### DIFF
--- a/.changeset/week-config-hash-2052.md
+++ b/.changeset/week-config-hash-2052.md
@@ -1,0 +1,8 @@
+---
+"@remnic/core": patch
+---
+
+Add `computeWeeklyConfigHash`: an order-stable digest of timezone, week start,
+and category definitions for weekly snapshot identity. The helper is exported
+from the activity timeline entry; caller wiring is a later slice.
+Part of #2052.
--- a/packages/remnic-core/src/activity/timeline/index.ts
+++ b/packages/remnic-core/src/activity/timeline/index.ts
@@ -16,6 +16,7 @@ export * from "./week-export.js";
 export * from "./week-snapshot.js";
 export * from "./week-previous.js";
 export * from "./week-recurring.js";
+export * from "./week-config-hash.js";
 export * from "./analysis-batch.js";
 export * from "./analysis-evidence.js";
 export * from "./analysis-prompt.js";
--- a/packages/remnic-core/src/activity/timeline/week-config-hash.test.ts
+++ b/packages/remnic-core/src/activity/timeline/week-config-hash.test.ts
@@ -1,0 +1,135 @@
+import test from "node:test";
+import assert from "node:assert/strict";
+
+import { computeWeeklyConfigHash, type WeeklyConfigInput } from "./week-config-hash.js";
+
+const BASE: WeeklyConfigInput = {
+  timezone: "Europe/Paris",
+  weekStartsOn: "monday",
+  categories: [
+    { id: "development", name: "Development" },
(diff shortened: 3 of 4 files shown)


joshuaswarren:
**Fixed in d24d79106.** Agreed on the mechanism: a non-blank check let `"Not/AZone"` or `"monady"` acquire a durable snapshot identity, and correcting the typo later would fork the stored history instead of continuing it.

The timezone now goes through the same `assertValidTimezone` the activity config and weekly builder use, and `weekStartsOn` is checked against a frozen weekday allow-list.

One note: `"utc"` is **not** a typo — IANA zone names are case-insensitive and `Intl.DateTimeFormat` accepts it, which I verified before writing the test. So my first draft's sample was wrong rather than the validator, and the case is now `"Chicago"` (a bare city with no area) alongside `"Not/AZone"` and `"America/Nowhere"`.

joshuaswarren:
**Round 2: one fixed in f0f187ea0, one declined with evidence.**

**Fixed — canonicalize before hashing.** Correct, and it is the same defect class this PR exists to close, one level up: `assertValidTimezone` accepts `utc` and `UTC`, and `US/Eastern` is an alias of `America/New_York`, so hashing the raw string gave two callers two snapshot identities for one zone. The digest now covers `Intl`'s resolved canonical zone. Regressions assert both equivalent pairs collide and that genuinely different zones still differ; they fail on the pre-fix tree.

**Declined — `"Chicago"` is not accepted on the supported runtime.** I checked before changing anything, on Node v22.23.1 (the version CI runs):

```
Chicago            rejected
America/Nowhere    rejected
Not/AZone          rejected
US/Eastern         ACCEPTED -> resolves to America/New_York
utc                ACCEPTED -> resolves to UTC
```

`Intl.DateTimeFormat` resolves IANA aliases and is case-insensitive, but it does not accept bare city names, so the test is correct as written. The canonicalization fix above makes the point moot anyway — `US/Eastern` and `utc` now hash identically to their canonical forms, which is what the alias behaviour actually calls for.

Issue #1957, opened 2026-07-16
research(shared-context): governed team memory - authority envelopes, staleness, contradictions, provenance + trajectory repository (2606.24535/MATM)


joshuaswarren:
## Summary

Formalize **governance for shared/team memory**: Remnic's `shared-context/` layer lets multiple agents exchange outputs, feedback, and priorities, but it currently has no explicit governance model. Add the four controls the recent literature identifies as the failure modes of fleet memory — (1) **origin + authority** on every shared item, (2) **staleness contracts** (TTL + supersession propagation), (3) **contradiction handling** across agents, and (4) **provenance preservation** through curation/synthesis steps. Also add a **trajectory repository**: completed agent task trajectories (already captured as causal trajectories) become shareable, reusable procedural artifacts with the same governance.

## Research grounding

- **Governed Shared Memory for Multi-Agent LLM Systems** (arXiv:2606.24535, 2026-06). Formalizes the fleet-memory problem; names the four failure modes: unauthorized leakage, stale propagation, contradiction persistence, provenance collapse.
- **Multi-Agent Transactive Memory (MATM)** (arXiv:2606.19911, 2026-06): population-level storage/retrieval of agent trajectories — trajectories encode reusable procedural knowledge but are typically discarded after one use.
- **Always-On Agents survey** (arXiv:2606.30306): six governance axes — authority, scope, mutability, provenance, recoverability, actionability.

## Current Remnic state

- `packages/remnic-core/src/shared-context/manager.ts` — write-output, feedback-record, priorities-append, cross-signals, daily curation. Items are markdown files with author + timestamp but no authority class, no TTL, no supersession, and curation (`curate_daily`) synthesizes without carrying per-claim provenance forward.
- `causal-trajectory.ts` / `causal-trajectory-graph.ts` — trajectories exist per-agent; procedural mining consumes them locally; nothing shares them.
- Compounding layer (weekly synthesize, promote candidate) has trust semantics for *promotion* but not for *cross-agent consumption*.

## Change

1. **Shared-item envelope** (`shared-context/governance.ts`): every shared-context write gains frontmatter `sharedBy` (agent id), `authority: "informational" | "advisory" | "binding"` (default `informational`; `binding` requires explicit flag), `expiresAt?` (ISO; default from config `sharedContextDefaultTtlDays`), `supersedes?` (item id). Legacy items without the envelope → `informational` + no expiry (least privilege, pattern 36).
[...]

PR #2762, opened 2026-08-20
feat(shared-context): resolve shared authority to the least privilege by default

joshuaswarren:
## Summary

Implements #1957's authority resolution: `informational | advisory | binding`, "default `informational`; `binding` requires explicit flag", and "legacy items without the envelope → `informational` + no expiry (least privilege, pattern 36)".

The security property is that nothing reaches `binding` by accident. Absent, non-string, and unrecognized authority values all resolve to `informational` — no case-folding, no trimming, so `"BINDING"` and `" binding"` are unrecognized data rather than near-misses that get normalized into privilege. A stored `"binding"` without the explicit opt-in **downgrades to advisory** rather than throwing, because a legacy item carrying that value is data, not a caller bug; the reasoning is in a comment at the branch.

`allowBinding` is the security switch, so a non-boolean throws instead of being coerced — a truthy `"false"` string must never enable binding.

The classes and type come from `governance.ts` (`SHARED_AUTHORITIES`) and are re-exported, not redefined, so there is one list. `compareSharedAuthority` is total: equal classes return 0 and an unknown class sorts as `informational` rather than throwing, so a mixed list stays sortable.

Part of #1957 (resolution helpers; read/write path wiring is a later slice, so the issue stays open).

## Test plan

- `NODE_OPTIONS=--conditions=remnic-source npx tsx --test packages/remnic-core/src/shared-context/authority-precedence.test.ts` — 10/10
- **Mutation-checked**: removing the `allowBinding` requirement so `binding` is honored unconditionally fails the flag test, so the security gate is genuinely asserted rather than incidentally true.

## Summary by CodeRabbit

* **New Features**
  * Added shared-authority resolution with clear precedence ordering.
  * Added least-privilege defaults for missing or invalid authority values.
  * Added safeguards requiring explicit approval for binding-level authority.
  * Added authority comparison and deterministic ordering support.
* **Bug Fixes**
  * Improved handling of unknown, invalid, and non-boolean authority inputs.


joshuaswarren:
**Fixed in efcd8ea4c**, by correcting the claim rather than adding an export.

You are right that the changeset read as announcing importable API. The module is source-internal like its shared-context siblings — `curation-lint.ts` and `supersession-chain.ts` are likewise absent from `src/index.ts` and the export map — so adding a subpath only for this one would make the family inconsistent. The changeset now states it is source-internal and that wiring callers is a later slice.

If the decision is that this family should be publicly exported, that is one follow-up covering all of them plus the `package.json` export map, not a one-off here.

joshuaswarren:
**Fixed in f62812fca.** Real and worth the Major: `input.allowBinding` was read at validation (line 36) and again at the decision (line 51), so a getter returning `undefined` then `true` resolved to `binding` with no validated opt-in — the switch bypassing itself.

It is now read once into a local, and both the validation and the decision use that value. The regression uses exactly that getter and additionally asserts the switch is read a single time, so re-splitting the read fails the test rather than silently restoring the window.

Issue #1952, opened 2026-07-16
research(memory): state-aware recall views - label current/historical/transition to kill ghost-memory failures (A-TMA)


joshuaswarren:
## Summary

Add **state-aware recall views** to the recall pipeline: label every recalled memory as `current`, `historical`, or `transition` relative to the query's implied time frame, and assemble "evidence packets" that keep superseded records visible *as history* instead of hiding or interleaving them. This directly targets the "ghost memory" failure class — old, current, and transition facts coexisting in injected context and misleading the agent.

## Research grounding

- **A-TMA: Decoupling State-Aware Memory Failures in Long-Term Agent Memory** (arXiv:2607.01935, 2026-07-02). Defines "ghost memory" failures; proposes a state-aware overlay that retains superseded records, builds evidence packets for the requested state view, and labels facts current/historical/transition. Its LTP benchmark shows current systems fail primarily on state confusion, not raw retrieval.
- **MINTEval** (arXiv:2605.18565): 27.9% average accuracy on multi-target interference over repeatedly-updated facts — the exact failure mode this issue mitigates.

## Current Remnic state

Remnic already has the storage-side half of this:
- `packages/remnic-core/src/temporal-supersession.ts` + `temporal-index.ts` — supersession links, `status: superseded`
- `asOf` historical recall pin (issue #680) on `remnic_recall`
- Lifecycle status filtering (AGENTS.md pattern 41: ACTIVE_STATUSES)

What's missing is the **read-side view layer**: recall today either filters superseded memories out entirely (default) or pins the whole corpus to a timestamp (`asOf`). There is no mode that returns the current fact *plus* its labeled predecessor chain when the query is about change ("when did X switch to Y", "what did we use before Z").

## Change

1. New module `packages/remnic-core/src/recall-state-view.ts`:
   - `annotateStateView(results, query, temporalIndex)` — tags each recall result `stateLabel: "current" | "historical" | "transition"` using the supersession chain from `temporal-index.ts`.
   - `transition` = a pair (superseded→superseding) where the query's intent is change-oriented (reuse `intent.ts` morphology-aware heuristics; add change-intent patterns: "when did", "used to", "before/after", "switched", "changed", conjugation variants per the intent guardrail).
[...]

PR #2763, opened 2026-08-20
feat(recall): check the state-view zero-diff compatibility promise

joshuaswarren:
## Summary

Makes #1952's compatibility promise checkable: "current items unchanged (**zero diff for existing users when no historical item qualifies**)".

That promise had no enforcement. The render path could gain a prefix, add a trailing space, or reorder a current-only result set and nothing would notice — the state-view feature would silently change injected context for users who never had a historical item to show.

`checkStateViewZeroDiff` compares the baseline injection against the annotated one and fails on any difference: `current_line_changed` names the offending memory id, and `order_changed` covers the case where every line matches but the sequence moved, since injected order drives what the agent reads first. Comparison is exact, so a renderer's trailing space is a diff.

The two ok reasons are deliberately distinguishable, documented in the module header: `unchanged` means the guarantee was checked and held, `no_historical_items` means it does not apply to this comparison. A caller can tell "verified" from "not applicable" instead of reading a bare true.

Labels come from the existing state-view type rather than a second list, and a non-`current` label in the baseline throws, since the baseline is by definition the pre-feature output.

Part of #1952 (guard helper; wiring it into a pipeline test is a later slice, so the issue stays open).

## Test plan

- `NODE_OPTIONS=--conditions=remnic-source npx tsx --test packages/remnic-core/src/recall-state-view-zero-diff.test.ts` — 11/11
- Covers identical lists, a changed line, a reorder, a trailing-space diff, the not-applicable path, a labelled baseline, unknown labels, duplicate ids, and input immutability.

## Summary by CodeRabbit

- **New Features**
  - Added validation for unchanged recall state-view text and memory ordering when only current items are present.
  - Reports whether validation is verified, not applicable, or identifies changed text or ordering.

- **Tests**
  - Added coverage for unchanged, modified, reordered, duplicate, invalid, and unsupported state-view data.
  - Added live-route checks confirming current entries remain unchanged when state views are enabled.


joshuaswarren:
**Fixed in cbcb75485.** This was the right call — a guard referenced only by its own unit test is machinery, and the repo's own scope rules say so.

Two cases now run the **real** `injectStateViewLines` pipeline with state views off and on, and assert the promise through the guard. One of them also asserts that the pipeline genuinely annotates a qualifying historical row (`[superseded 2026-08-01 by m-1]`), so the zero-diff case cannot pass merely because rendering happens to be a no-op everywhere — that sanity assertion is what makes the other one meaningful.

While wiring it I also renamed the two ok reasons: `unchanged` meant "nothing was compared" and `no_historical_items` meant "checked and held", which is backwards from how both read. They are now `verified` and `not_applicable`, with the old names noted in the module header so the inversion is not reintroduced.

joshuaswarren:
**Fixed in 3c6fe3db0 — and you were right for a worse reason than the one you gave.**

Rewriting the test against the live route (`applyRecallStateViews` → `RecallResultFormatter.formatQmdResultEntries`, as `orchestration/recall-entry.ts` calls them) exposed that my fixture was **vacuous**. `parseConfig` deliberately does not carry `recallStateViews` — `recall-state-view-wire.ts` says "parseConfig cannot grow. Operators set `recallStateViews` on the live PluginConfig extra field" — so passing the flag *into* `parseConfig` produced a disabled render, and the zero-diff comparison was between two identical baselines. It would have passed no matter what the enabled path did.

The flag is now set the way production sets it, and a dedicated test asserts both halves of the enable condition (the flag is absent from `parseConfig`, present on the live object, and the query is change-oriented) so the vacuum cannot come back unnoticed.

Verified by mutation, not by assertion:

| mutation on the enabled path | zero-diff test | payload test |
|---|---|---|
| prefix every row (`[sv] …`) | **fails** | passes |
| drop the second row | **fails** | **fails** |
| reorder entries in the formatter (affects both paths) | passes | **fails** |

The third row is deliberate: a formatter change that perturbs baseline and annotated equally is a formatter regression, not a state-views compatibility break, and this guard should not claim to catch it.

One thing I am **not** doing here: `injectStateViewLines`/`renderStateViewLine` landed in an earlier slice and, as you note, have no production caller. Deleting them belongs in its own change, not in a PR about the guard.

PR #2764, opened 2026-08-20
chore(ci): gate link vocabularies, batch thread requests, document three defect classes

joshuaswarren:
## Summary

Five process fixes for failures this run actually hit. A sixth candidate was **measured and rejected** rather than shipped as noise.

### Shipped

1. **`scripts/check-nav-link-vocabularies.mjs`** (wired into the `checks` job) fails when the shared navigation parser does not know a link type the stepper accepts or that is actually persisted. Three lists had drifted apart silently: the parser rejected `follows`/`references`/`related` — which real frontmatter carries, so a traversal over stored links dropped those neighbors — and `supersedes`, which the stepper accepts. It took **two review rounds** to find. The unit test replays that exact pre-fix divergence and asserts each missing type is attributed to the list that requires it.
2. **`scripts/pr-review-threads.mjs`** issues ONE aliased query for every PR and ONE aliased mutation for every thread, instead of 2N requests. The naive per-PR loop exhausted the GraphQL point budget **three times in this run**, each time leaving `unresolved-review-threads` red until the budget refilled — roughly two and a half hours of pure waiting. Smoke-tested live against a real PR.
3. **`AGENTS.md` pattern 46** — enumerate keys with `getOwnPropertyNames`, read them with `hasOwn`, and treat present-with-`undefined` as invalid rather than absent. Two allow-list bypasses shipped one review round apart on exactly this mismatch (inherited property, then non-enumerable property).
4. **`AGENTS.md` pattern 47** — never annotate a frozen `as const` array `readonly T[]`; it widens the derived literal union back to `string`. This shipped as a regression *inside a hardening fix*. Includes the `@ts-expect-error` pin that makes a future widening fail the build, since an unused directive is itself an error.
5. **`AGENTS.md` pattern 48** — every claim in a PR body, module header, or changeset must name the code or test enforcing it, or be weakened to what is true. Two false claims were caught by reviewers this run ("structurally incapable of carrying content" on a validator that accepted prose; "internal helper" on an exported module).

### Measured and rejected

Per the README ground rule — run a condition against the tree before shipping it:

- A regex for the pattern-47 widening (`: readonly T[] = Object.freeze([...])`) matches **13 legitimate frozen data arrays** (seed vocabularies, prose frames, fixture memories) where no `(typeof X)[number]` union is derived and the annotation is the intended type. Separating them needs two facts in different places, so it is not a single-line signature. Receipt recorded in `.omp/rules/README.md`; the remedy lives in pattern 47.

## Test plan

- `node scripts/check-nav-link-vocabularies.mjs` — passes on current `main` (parser knows 9 types, covering 5 stepper and 5 persisted), locking in the batch-9 fix.
- `node --test tests/check-nav-link-vocabularies.test.mjs tests/pr-review-threads.test.mjs` — 11/11.
- `node scripts/pr-review-threads.mjs list joshuaswarren/remnic 2756` — exercised against the live API in a single request.

## Summary by CodeRabbit

* **New Features**
  * Added a command-line tool to list unresolved pull request review threads and resolve multiple threads in one operation.
  * Added pagination, duplicate handling, input validation, and clear error reporting for review-thread operations.

* **Bug Fixes**
  * Added a CI check to detect missing navigation link types and vocabulary inconsistencies.
  * Improved vocabulary checks to ignore commented-out entries while preserving URLs and other quoted text.

* **Documentation**
  * Expanded development guidance for safer validation, preserving type information, and enforcing documented guarantees.

### DIFF
--- a/.changeset/ci-b10-hardening.md
+++ b/.changeset/ci-b10-hardening.md
@@ -2,4 +2,4 @@
 "@remnic/core": patch
 ---
 
-Add a CI gate asserting the navigation link-type vocabularies agree, a batched review-thread helper that replaces 2N GraphQL requests with two, and three review-prevention patterns covering key-validation bypasses, frozen-const type widening, and unenforced claims.
+Add a CI gate asserting the navigation link-type vocabularies agree (`scripts/check-nav-link-vocabularies.mjs`, wired into the `checks` job), a batched review-thread helper that replaces per-PR list queries and per-thread resolve mutations with one aliased query and one aliased mutation — plus one extra query per additional page for a PR with more than 50 threads — and three review defect classes documented in `.omp/rules/README.md`. The proposed rule for literal-union widening was measured against the tree, matched 13 legitimate frozen data arrays, and is recorded as rejected rather than shipped.
--- a/.changeset/ci-b10-hardening.md
+++ b/.changeset/ci-b10-hardening.md
@@ -1,5 +1,0 @@
----
-"@remnic/core": patch
----
-
-Add a CI gate asserting the navigation link-type vocabularies agree (`scripts/check-nav-link-vocabularies.mjs`, wired into the `checks` job), a batched review-thread helper that replaces per-PR list queries and per-thread resolve mutations with one aliased query and one aliased mutation — plus one extra query per additional page for a PR with more than 50 threads — and three review defect classes documented in `.omp/rules/README.md`. The proposed rule for literal-union widening was measured against the tree, matched 13 legitimate frozen data arrays, and is recorded as rejected rather than shipped.
--- a/.changeset/ci-b10-hardening.md
+++ b/.changeset/ci-b10-hardening.md
@@ -1,0 +1,5 @@
+---
+"@remnic/core": patch
+---
+
+Add a CI gate asserting the navigation link-type vocabularies agree, a batched review-thread helper that replaces 2N GraphQL requests with two, and three review-prevention patterns covering key-validation bypasses, frozen-const type widening, and unenforced claims.
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -180,6 +180,13 @@ jobs:
         env:
           REMNIC_REGEX_SAFETY_BASE_REF: ${{ github.event_name == 'pull_request' && format('origin/{0}', github.event.pull_request.base.ref) || 'origin/main' }}
 
+      # Navigation link-type vocabulary agreement (#1956 review fallout): the
+      # shared parser must know every link type the stepper accepts and every
+      # type that is actually persisted. Three lists drifted apart silently and
+      # cost two review rounds to find.
+      - name: Navigation link vocabularies agree
+        run: node scripts/check-nav-link-vocabularies.mjs
+
   # The root/misc suite runs on a hosted runner. Package tests use two
   # deterministic file shards on the self-hosted pool when it is safe.
   # selectTestShard is an exact partition (enforced by test), so the package
--- a/.omp/rules/README.md
+++ b/.omp/rules/README.md
@@ -88,6 +88,16 @@ conditions without new evidence; each baseline below is the receipt.
   where *any* rejection is precisely the contract (`readFile` on a
   deleted path, a loop of clearly-typed invalid inputs). Tightening the
   matcher is a review judgement, not a signature.
+- **`: readonly T[] = Object.freeze([...])`** (literal-union widening) — the
+  defect is real and shipped once as a regression inside a hardening fix,
+  but the signature matches 13 legitimate frozen *data* arrays (seed
+  vocabularies, prose frames, fixture memories) where no
+  `(typeof X)[number]` union is derived and the annotation is exactly the
+  intended type. Distinguishing them needs two facts in different places —
+  an `as const` initializer AND a derived union — so it is not a
+  single-line signature. `AGENTS.md` pattern 47 carries the remedy,
+  including the `@ts-expect-error` pin that makes a future widening fail
+  the build.
 
 The dominant review clusters in the 2026-06-20..07-04 subset (605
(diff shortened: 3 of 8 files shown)


joshuaswarren:
**Round 1 fixes landed in ff8e61b7a.** Both findings were real bugs in the helper this PR adds, and one of them defeated its purpose.

- **Pagination.** A PR with more than 50 threads (resolved ones count toward the page) silently truncated, so the tool could print no unresolved threads while the guard stayed red — precisely the failure it exists to prevent. Queries now carry `pageInfo`, and only the PRs that report another page are revisited, so the common case is still one request. Tests cover the cursor threading and that a completed PR is not requeried.
- **Duplicate-id exit status.** `buildResolveMutation` deduplicates by design, so comparing against the caller's raw `ids.length` failed a run that resolved everything. It now measures against the unique ids actually sent, with a test.

A third fix came from smoke-testing it: a rate limit produced a raw `execFileSync` stack trace — the one condition this helper exists to survive. It now reports a single explanatory line, and any GraphQL `errors` payload surfaces as a message instead of parsing through as success. Verified against a live rate-limited call.

joshuaswarren:
**Round 2: all six fixed in 7be9ad62c.** Every one was real, and two were mine claiming things the code did not do.

- **GraphQL injection (Major).** Correct on the mechanism: `execFileSync` passes argv without a shell, so the exposure is the *operation* changing under the configured `gh` credential. Owner, repo, cursors, and thread ids are now charset-checked (`[A-Za-z0-9._-]+` and `[A-Za-z0-9_=-]+`) and serialized as GraphQL string literals. A test feeds the cursor `x" ) { evil }` and asserts the quote escapes in place and no second `reviewThreads(` selection appears.
- **stderr diagnostics.** Fixed — both bots caught this. `gh` writes the concise diagnostic to stderr and the JSON body to stdout; both are inspected before classifying, and the generic failure reports stderr first.
- **The documented pipe never worked.** Also caught by both. `resolve` read only argv, so `list | resolve` always threw the usage error. It now falls back to stdin, with a test over real `list` output; the header documents both forms.
- **Pagination cap.** Removed. A valid cursor chain has to be allowed to finish however long it is; the terminating condition is now a cursor that repeats for the same PR, which is the actual loop.
- **Comments as vocabulary.** Fixed with a string-aware comment strip ahead of both declaration readers, so `// "related"` no longer reads as supported while a `//` inside a string literal survives. Tests cover both. Not a TypeScript AST: the gate reads two `const … as const` arrays and one union, and a tokenizer would be more machinery than the check it guards.
- **Changeset claim.** Corrected — it now says one query plus one extra per additional page, instead of a flat 2N-to-2.

joshuaswarren:
**Fixed in d2b5fcbab.** Right, and it is the healthy path that was broken: when every queried PR is already clean, `list` emits nothing, so the documented pipeline exited 1 on success.

An empty pipe and no pipe are now distinct states. `threadIdsFromStdin` returns `[]` when a pipe carried no ids and `null` when stdin is a TTY, so `list | resolve` on a clean repo prints `nothing to resolve` and exits 0 while a bare interactive `resolve` still fails with the usage line.

One thing your report did not cover that I hit while testing it: an empty pipe can raise `EAGAIN` on `readFileSync(0)`, which the previous code caught and turned into `""` — indistinguishable from "no pipe". That is now treated as the empty pipe it is, otherwise the fix would only have worked when the producer left the pipe open.

Verified end to end against the live API: `list joshuaswarren/remnic 2760 | resolve` → `nothing to resolve`, exit 0.

joshuaswarren:
**Both fixed in d02aee02a.** Both were silent-wrong-answer shapes, which is the class worth spending a round on.

**Empty pipeline needed proof, not an assumption.** You are right that a `list` which died early — rate limit, inaccessible PR — produces the same empty stdin as a genuinely clean run, and without `pipefail` the pipeline then reports success having queried nothing. That is worse than the bug I was fixing.

`list` now prints an end-of-list marker after its final row, and `resolve` requires that marker before accepting empty input. Absent, it fails with "no evidence the queried PRs are clean". I hit exactly your scenario while testing it: the GraphQL budget was exhausted, `list` died, and the pipeline correctly refused instead of reporting clean.

**Non-decimal PR tokens.** `Number("0x10")` is 16 and `Number("1e2")` is 100, so the helper would report another PR's thread state as the one requested. PR numbers are now validated as decimal digits on the original token before conversion, with the boundary cases pinned (`0x10`, `1e2`, `12.0`, `+12`, `" 12"`, `0`).

Verified: sentinel-only stdin exits 0 with "list finished with no unresolved threads"; empty stdin without the marker fails; `list owner/repo 0x10` is refused.

joshuaswarren:
**Both fixed in 01104c54d.**

**The gate could pass while the invariant was broken.** Correct, and that is disqualifying for a gate: it extracted string literals and silently ignored anything else, so a union composed through an alias or a spread would sail through. It now fails closed and names what it cannot read. Verified against the real declaration rather than a fixture — injecting `...EXTRA_TYPES` into `NAVIGATE_LINK_TYPES` produces:

```
check-nav-link-vocabularies: NAVIGATE_LINK_TYPES in packages/remnic-core/src/recall-navigate-link.ts
has non-literal member(s) [...EXTRA_TYPES] this gate cannot read. Extend the extractor rather than
letting the vocabulary check pass on a partial list.
```

**The changeset was wrong twice, so it is deleted.** It targeted `@remnic/core` while this PR changes no file under `packages/` — only root `scripts/`, `tests/`, workflows and docs — which would have published a phantom core patch. It also sent readers to `.omp/rules/README.md` for three defect classes that actually live in AGENTS.md as patterns 46–48.

The root workspace is private and no gate requires a changeset for tooling-only changes, so deleting it is the honest fix rather than inventing a target for a release containing no package code. Worth noting for the earlier CI PRs in this series: they carry the same phantom `@remnic/core` changeset, and they are already merged. That is a real (small) hygiene problem I introduced, and it belongs in a follow-up sweep rather than here.

PR #2765, opened 2026-08-20
chore(deps): close every open Dependabot alert

joshuaswarren:
## Summary

Closes every open Dependabot alert.

| package | before | after | alerts |
|---|---|---|---|
| `hono` | 4.12.29 | 4.13.3 | 4 (one low, three medium) |
| `@hono/node-server` | 1.19.14 | 1.19.15 | 1 medium |
| `postcss` | 8.5.17 | 8.5.26 | 2 (one **high**) |
| `tar` | 7.5.20 | 7.5.22 | 1 medium |
| `torch` (pip) | 2.12.1 | 2.13.0+cu126 | 1 low |

All four npm packages are transitive. `@hono/node-server` needed a
`pnpm-workspace.yaml` override because the `@modelcontextprotocol/sdk` range is
`^1.19.9 || ^2.0.5`, so a plain resolver update jumps to the 2.x major; the
override pins the patched 1.x instead. The existing `overrides:` block was
extended rather than a second mechanism introduced.

`torch` follows the file's existing `==<version>+cu126` pin style.

## Verification

- `node scripts/pnpm.mjs install --frozen-lockfile` succeeds.
- Lockfile resolves `hono@4.13.3`, `@hono/node-server@1.19.15`,
  `postcss@8.5.26`, `tar@7.5.22` — each at or above its first patched version.
- Lockfile regenerated by the pinned wrapper (`pnpm@10.32.1`), not hand-edited.
- No unrelated dependency moved; `nanoid` 3.3.15 -> 3.3.18 rides along as
  postcss's own dependency.

### DIFF
--- a/.changeset/dep-security-bumps.md
+++ b/.changeset/dep-security-bumps.md
@@ -1,0 +1,6 @@
+---
+"@remnic/core": patch
+"@remnic/bench": patch
+---
+
+Security bumps for open Dependabot alerts: hono 4.12.29 → 4.13.3, @hono/node-server 1.19.14 → 1.19.15 (held on the patched 1.x line by a pnpm-workspace override because the sdk range otherwise resolves the 2.x major), postcss 8.5.16 → 8.5.26 (dev-only, via tsup), tar 7.5.19 → 7.5.22 (via node-gyp), and the model-lab torch pin raised to 2.13.0+cu126.
--- a/model-lab/requirements.txt
+++ b/model-lab/requirements.txt
@@ -31,7 +31,7 @@
 #     deserialization cluster GHSA-qxrp-vhvm-j765 / -hxxf-235m-72v3 /
 #     -wrfc-pvp9-mr9g (all < 4.48.0).
 --extra-index-url https://download.pytorch.org/whl/cu126
-torch==2.12.1+cu126
+torch==2.13.0+cu126
 transformers==5.5.0
 datasets==3.6.0
 huggingface-hub==1.5.0  # transformers 5.5.0 requires huggingface-hub>=1.5.0,<2.0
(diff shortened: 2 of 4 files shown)


joshuaswarren:
**Declined — the wheel exists.** I checked the index before changing anything:

```
$ curl -sL https://download.pytorch.org/whl/cu126/torch/ | rg -o 'torch-2\.13\.0\+cu126-cp3(11|12|13)[^"]*\.whl' | sort -u
torch-2.13.0+cu126-cp311-cp311-manylinux_2_28_aarch64.whl
torch-2.13.0+cu126-cp311-cp311-manylinux_2_28_x86_64.whl
torch-2.13.0+cu126-cp311-cp311-win_amd64.whl
torch-2.13.0+cu126-cp312-cp312-manylinux_2_28_aarch64.whl
torch-2.13.0+cu126-cp312-cp312-manylinux_2_28_x86_64.whl
torch-2.13.0+cu126-cp312-cp312-win_amd64.whl
torch-2.13.0+cu126-cp313-cp313-manylinux_2_28_aarch64.whl
torch-2.13.0+cu126-cp313-cp313-manylinux_2_28_x86_64.whl
torch-2.13.0+cu126-cp313-cp313-win_amd64.whl
```

Nine wheels are published on the cu126 index. `model-lab/setup.sh` refuses
anything below Python 3.12 (repo standard), and cp312 and cp313 wheels are both
present for linux x86_64 and aarch64, so every interpreter this tree supports
has a wheel.

Worth noting for anyone reading later: 2.13.0 dropped the cp310 wheel that
2.12.0 had. That would matter if the bootstrap allowed 3.10 — it does not, and
the version gate in `setup.sh` is what makes this pin safe.

PR #2766, opened 2026-08-20
fix(security): clear all five CodeQL alerts

joshuaswarren:
## Summary

Clears all five open CodeQL alerts.

### js/clear-text-logging (HIGH) — #223, #220

`parseConfig` error strings embed raw config values: `resolveEnvVars` throws
`Malformed environment variable placeholder: <the value>` for `openaiApiKey`
and `localLlmApiKey`, and several rejects stringify the raw field. The
catch-all `console.error(err.message)` in the `codegraph` and `export-okf`
commands could therefore print key material.

Config loading now runs in its own try block, and on failure the commands print
the error class plus a config dump passed through a new `redactConfigForLog`.
Redaction is **by key name**, applied recursively, so a newly added
`anthropicApiKey` is masked by construction rather than by someone remembering
to add a pattern — a test pins exactly that case with a key the redactor was
never told about. The pattern mirrors `SENSITIVE_KEY_RE` in core's
admin-surfaces, the existing in-repo convention.

### js/polynomial-redos (HIGH) — #222, #219, #195

| site | defect | fix |
|---|---|---|
| `reconcile/cursor.ts` | `/\/+$/` backtracks quadratically on long slash runs that do not end in a slash | `stripTrailingSlashes`, a bounded character loop |
| `transfer/export-okf.ts` | `/^#\s+(.+)$/` — `\s+` and `.+` overlap, so a failed match backtracks over every split point | `/^#\s+(\S.*)$/`, disjoint capture boundary |
| `location/matching.ts` | `/^-+\|-+$/` quadratic on dash runs in isolation | single-character strips, which the preceding `/[^a-z0-9]+/` collapse makes complete |

## Verification

Discrimination measured per site with file-level restores (not `git stash` —
it is repo-global across worktrees):

- `cursor.ts` reverted → its timing test fails.
- `export-okf.ts` reverted with the `firstHeading` export kept → its timing
  test fails.
- `matching.ts`: a timing test could **not** discriminate, because the collapse
  already bounds the trim's input, so it was replaced with a test of the
  invariant the strips depend on (no edge may hold two dashes). Weakening the
  collapse to `/[^a-z0-9]/` fails it.

Suites: `matching` 4/4, `cursor` 12/12, `export-okf` 11/11, `redact-config`
3/3. `tsc --noEmit -p packages/remnic-core` clean; the CLI's remaining
resolution errors are the unbuilt-workspace artifact a fresh worktree always
shows (127 of them on clean `main` too), not from this change.

## Summary by CodeRabbit

* **New Features**
  * Configuration diagnostics show detected keys and unresolved placeholders without exposing sensitive values.
  * Reports indicate when scanning stops early or reaches the key limit.
  * Empty, malformed, and whitespace-only configuration files are handled more reliably.

* **Bug Fixes**
  * Improved configuration loading and error reporting for code graph and OKF export commands.

* **Performance**
  * Improved processing of unusually long inputs, including location tags, peer identifiers, headings, and configuration text.

* **Tests**
  * Expanded coverage for redaction, parsing edge cases, Unicode, duplicate keys, and large inputs.


joshuaswarren:
**Round 1 fixes in 79492e707. CodeQL re-flagging its own fix was the important signal, and it was right.**

**The two clear-text-logging alerts (#225, #226).** Redaction did not close them, and on reflection it should not have: taint analysis cannot see through a generic `redact<T>(config)`, and more to the point a redactor is only as complete as its key pattern's last edit. That is a maintenance promise, not a guarantee.

The failure path now reports the config's **shape** — key paths, value kinds, and which keys hold an unresolved `${...}` placeholder, which is the failure `resolveEnvVars` actually raises. No value is read into the output at all, so there is no pattern to keep current and a newly named secret field is safe by construction.

The diagnosis got *better*, not worse. Where an operator previously saw a dump with one field masked, they now see the failing key named:

```
codegraph export-okf: failed to load config at /path/openclaw.json (RangeError)
  localLlmApiKey: string
  memoryDir: string
  openaiApiKey: string (unresolved ${...} placeholder)
  qmd: object
  qmd.collection: string
```

Tests assert that no value of any kind reaches the output — including a secret-named key the module was never told about, and the placeholder body itself, which is a value too — while paths and kinds survive, ordering is deterministic, and the entry cap holds.

**The unsound generic return.** Correct, and it went away with the module: `redact-config.ts` is deleted rather than left beside its replacement. `describeConfigShape` returns `ConfigShapeEntry[]`, which is what it actually produces.

joshuaswarren:
**Round 2: the leak you both found was real, and the fix is in 144ee55ed.**

You were right and it was the worst kind of wrong on my part: this code path exists *because* the file did not parse, and I reached for a regex, which has no structural context. `{"openaiApiKey":"secret":}` puts a colon after the **value**, so the value matched the key pattern and got printed — a credential leak inside the clear-text-logging fix. That is the third version of the same mistake in this PR (redactor → shape walk → regex), each one still touching data it had no business touching.

It is now a bounded positional scanner:

- A string is reported only when the scanner is genuinely in an object-key slot. Array elements and value-position strings are never reported.
- A key slot holding anything but a string **stops the scan**. That is what closes `{"a":1, /* "secret": */ }` — the case that still leaked in my first attempt at this, because after the comma the scanner was expecting a key and happily read the string inside the comment.
- No value is ever copied. The only thing read from one is whether it has the `${...}` shape.

Also fixed from your other two comments: `unresolved` is bounded to reported keys (it could otherwise grow past `maxKeys` and make formatting `maxKeys × unresolved` comparisons), and a new `truncated` flag tells an operator the scan stopped early rather than implying the file held no more keys.

Tests: both adversarial inputs you named, arrays, unterminated strings, the cap, an out-of-charset key (no phantom fragment), a 20k-token malformed file under a 1s budget, and an 18-input sweep asserting no value-position secret ever prints. A secret-shaped *key name* is reported by design — that is a field name, not a credential.

On the stale gate report: `check-regex-safety.mjs` did fail at the head you looked at, and it was my own new regex that tripped it — correctly, since it was the same nested-quantifier shape this PR removes elsewhere. Fixed in d88d8f66f before this rewrite, and the regexes are gone entirely now. The gate passes.

joshuaswarren:
**Both fixed in 55f3d950a. The first one was a real leak and your repro was exact.**

**Invalid scalar, resynced structure.** I reproduced it before fixing:

```
input:  {"openaiApiKey":oops<200 spaces>{"value-that-must-never-print":0}}
before: LEAKS: true   ->   openaiApiKey | value-that-must-never-print
```

Skipping unknown characters one at a time let the later `{` look like a genuine nested object, so the next value-position string was read as a key and printed. And as you note, with enough padding the `JSON.parse` message does not even mention that region, so this path was the only thing between that value and stderr.

Scalars are now read as whole tokens and validated against JSON's grammar; an invalid one stops the scan, because past an invalid token there is no structure left to trust. The regression is your input verbatim, and valid scalars (`-2.5`, `1e3`, `true`, `false`, `null`) still scan through.

That is the fourth iteration of this fix, and the pattern across all four is the same: every time I let the reporter keep *interpreting* broken input, it found a way to promote a value into a key. Stopping at the first thing that cannot be trusted is the only version that holds.

**Multi-character escapes.** Correct, and correctly triaged as diagnostic-only — the scanner never echoes values. A fixed `+2` left the four hex digits of `\uXXXX` in the string, the key then failed the charset test, and it vanished from the report entirely. Escapes now consume their full width, so a unicode-escaped key is reported rather than silently dropped. A dropped key is still a lie to the operator, so it was worth the three lines.

